PR #7, opened 2023-07-06
Minor changes to the API

jcpitre:
- Added legacy_id
- Changed syntax of /feeds/gtfs/{id}/datasets to /feeds/{id}/datasets/gtfs (moved gtfs at the end)
- Added a /datasets endpoint to get a list of basic dataset data regardless of type.
- Reduced the number of common fields returned when obtaining a generic list of feeds of datasets.
- Added /datasets/{id} endpoint as per PR comment.

### DIFF
--- a/docs/DatabaseCatalogAPI.yaml
+++ b/docs/DatabaseCatalogAPI.yaml
@@ -225,37 +225,39 @@ components:
           description: Unique identifier used as a key for the feeds table.
           type: string
           example: feed_0
-        legacy_id:
-          description: An optional legacy ID that could be useful for data migration.
-          type: string
-          example: ID_101
         data_type:
           $ref: "#/components/schemas/DataType"
         status:
           $ref: "#/components/schemas/ElementStatus"
-        provider:
-          description: A commonly used name for the transit provider included in the feed.
-          type: string
-          example: London Transit Commission
-        feed_name:
-          description: >
-            An optional description of the data feed, e.g to specify if the data feed is an aggregate of 
-            multiple providers, or which network is represented by the feed.
-          type: string
-          example: Atlantic Station Shuttle (FREE RIDE)
-        note:
-          description: A note to clarify complex use cases for consumers.
-          type: string
+
 
     GtfsFeed:
       allOf:
         - $ref: "#/components/schemas/BasicFeed"
         - type: object
           properties:
+            legacy_id:
+                description: An optional legacy ID that could be useful for data migration.
+                type: string
+                example: ID_101
+            provider:
+              description: A commonly used name for the transit provider included in the feed.
+              type: string
+              example: London Transit Commission
+            feed_name:
+              description: >
+                An optional description of the data feed, e.g to specify if the data feed is an aggregate of 
+                multiple providers, or which network is represented by the feed.
+              type: string
+              example: Atlantic Station Shuttle (FREE RIDE)
+            note:
+              description: A note to clarify complex use cases for consumers.
+              type: string
             location:
               $ref: "#/components/schemas/FeedLocation"
             urls:
               $ref: "#/components/schemas/FeedUrls"
+
 
     BasicFeeds:
       type: array
@@ -363,10 +365,6 @@ components:
           description: Unique identifier used as a key for the datasets table.
           type: string
           example: dataset_0
-        legacy_id:
-          description: An optional legacy ID that could be useful for data migration.
-          type: string
-          example: ID_101
         feed_id:
           description: ID of the feed related to this dataset.
           type: string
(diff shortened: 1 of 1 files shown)


docs/DatabaseCatalogAPI.yaml
```diff
@@ -311,13 +341,13 @@ components:
...
 
-    FeedStatus:
+    ElementStatus:
       description: > 
-        Describes status of the feed. Should be one of 
-          * `active` Feed should be used in public trip planners.
-          * `deprecated` Feed is explicitly deprecated and should not be used in public trip planners.
-          * `inactive` Feed hasn't been recently updated and should be used at risk of providing outdated information.
-          * `development` Feed is being used for development purposes and should not be used in public trip planners.
+        Describes status of the element (Feed or Dataset). Should be one of
```
emmambd, line 346: Can you explain why status would apply to both dataset + feed? I think this is a feed level property.

docs/DatabaseCatalogAPI.yaml
```diff
@@ -199,10 +225,14 @@ components:
           description: Unique identifier used as a key for the feeds table.
           type: string
           example: feed_0
+        legacy_id:
```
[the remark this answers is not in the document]
jcpitre: I think the issue of legacy ids should be subject of some use cases research. Can I leave it as is for now?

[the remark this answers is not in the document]
emmambd: Absolutely! I'll make an issue for it to dig in further.

docs/DatabaseCatalogAPI.yaml
```diff
@@ -160,6 +160,26 @@ paths:
...
+      tags:
+        - "datasets"
+
+      security:
+        - ApiKeyAuth: []
+      responses:
+        200:
+          description: >
+            Successful pull of the dataset common info for the provided ID.
+            This info has a reduced set of fields that are common to all types of dataset.
```
emmambd, line 177: Ideally this includes all fields specific to the data type. If it's a reduced set of fields, I don't think it has much value to include, and it makes more sense to just use the data type specific end point, e.g `/datasets/gtfs/{id}` Thoughts? Having a reduced set of fields for `/datasets` and `/feeds` makes sense to me since the response will be huge, but not for id specific requests. cc @jcpitre @davidgamez

docs/DatabaseCatalogAPI.yaml
```diff
@@ -324,58 +376,68 @@ components:
...
+              type: string
+              example: https://storage.googleapis.com/storage/v1/b/mdb-latest/o/us-maine-casco-bay-lines-gtfs-1.zip?alt=media
+            note:
+              description: A note to clarify complex use cases for consumers.
+              type: string
+            download_date:
+              type: string
+              example: 2020-12-31
+              format: date
+            creation_date:
```
emmambd, line 415: Tagging @davidgamez since @jcpitre mentioned you suggested this: 

Is there ever a situation where the `creation_date` for a dataset will be different from our `download date`? If we're not extracting the actual feed and looking at `feed_info.txt`, `calendar.txt` or calendar_dates.txt, I assume this'll always line up with when we compare datasets and download the new one.

docs/DatabaseCatalogAPI.yaml
jcpitre: I  believe the rest api best practices suggests that the data returned from a given endpoint should always use the same schema. It's not an obligation but I think we would need a better reason to stray from that.
One case is if for some reason you have the ID of a feed or dataset, but nothing else. You will then be able to get the reduced feed/dataset that will give you the type, then you can get the full data. A bit convoluted maybe, but it's a corner case IMO. 
Maybe in the spirit of an MVP we could get rid of /feeds/{id} and /datasets/{id} altogether. If the case described presents itself, the workaround would be to get the whole list then keep only the one with the requested ID.

emmambd: Got it! I understand now the need for a feed type specific end point given the rest API best practices and the processing required. Thanks for educating me @jcpitre @davidgamez on that. :smile:

docs/DatabaseCatalogAPI.yaml
davidgamez: You are right; as for MVP, we are not parsing any date from the GTFS spec itself. What about if we make it simple for consumers and collapse `download_date`, `creation_date `, and `last_update_date ` on one field?  Keeping only the `date` that MobilityData stores the file.

emmambd: This makes sense to me for datasets! `download_date` is the clearest to me. (`date` might not be specific enough in the future as we add to the MVP.) Any other thoughts on naming?

davidgamez: +1 for `download_date`.

Issue #21, opened 2023-07-07
Infrastructure setup for API: Proof of concept


emmambd:
## Acceptance Criteria

- [ ] Proof of concept API endpoints on GCP 
- [ ] Initial CI pipeline setup 
- [x] Document the local development process
- [x] Should be able to do local development without asking for access from MobilityData

This issue is part of #2

PR #24, opened 2023-07-11
feat: Generates api stub from Openapi schema

davidgamez:
## Description

This PR adds the minimum API implementation using [OpenApi Python FastAPI generator](https://openapi-generator.tech/docs/generators/python-fastapi)

Part of #21

## Tasks Before Ready

- [x]  Fix requirements installation for Python 3.11
- [ ]  Investigate imports from the main project folder
- [x] Fix broken tests(docker depends on tests)
- [x] Fix docker implementation
- [x] Document code generation workflow
- [x] Review FastAPI and default parameters' value


README.md
```diff
@@ -9,3 +9,69 @@ Mobility Feed API is not released yet; any code or service hosted is considered
...
+```bash
+scripts/api-tests.sh
+```
+
+To run a single test file:
+```bash
+scripts/api-tests.sh <my_test_filename>.py
+```
+
+## Running with Docker
```
cka-y, line 71: We should probably specify here that the generation of feed_gen is necessary before running Docker. Or we could add the generation in the Docker file itself

cka-y (approved):

README.md
davidgamez: That's a good point. Let's add it as part of the docker build phase to make sure the module is present.

davidgamez: I end up documenting enhancing the documentation of the `docker-compose` as an easier solution.

Issue #51, opened 2023-07-19
Develop metadata endpoint


emmambd:
## User Stories

-  I need to know which version of the API I'm using in my app, so I can debug

## Acceptance Criteria

- Should return the API metadata as specified in [Swagger](https://mobilitydata.github.io/mobility-feed-api/SwaggerUI/index.html#/metadata/get_metadata)

Issue #35, opened 2023-07-17
Develop feeds/{id} endpoint


emmambd:
## User Stories

- I want to fetch feeds based on their ID, so I can consume the fields easily and use them in my app.

## Acceptance Criteria

- When the ID is inputted, should return 1 feed with all fields that are common to all feed types. [Example here](https://mobilitydata.github.io/mobility-feed-api/SwaggerUI/index.html#/feeds/get_feeds__id_)

Issue #32, opened 2023-07-17
Develop feeds endpoint


emmambd:
## User Stories

- I want to fetch the full list of feeds from the Mobility Database, so I can get familiar with the schema and quickly discover what types of data are available. 
- I want to filter the feed list by status, so I can pull only the active feeds, test feeds in development, or check and update the inactive feeds in my app. 
- I want to search for specific providers in the feed list, so I can find the feeds I need to add, compare or analyze.
- I want to search for specific producer URLs, so I can search for feeds I already know of or have in my system easily.

## Acceptance Criteria

- Should return a list of feeds with all the fields that are common to all feed types, including when no parameters are specified [Example here](https://mobilitydata.github.io/mobility-feed-api/SwaggerUI/index.html#/feeds/get_feeds)

- Should be able to send the following optional parameters [(specified in Swagger)](https://mobilitydata.github.io/mobility-feed-api/SwaggerUI/index.html#/feeds/get_feeds): 

   - int `limit` 
   - int `offset`
   - string `filter`
      - filter options
          - enum `status`: options `active`, `inactive`, `development`, `deprecated`
              - feeds with an empty status field should be returned when the `status=active` 
          - contains string `provider`, e.g `"Kearns"`
          - contains string `producer_URL` e.g `"download.gtfs.de/germany/nv_free"`
   - string `sort`
         - sort options 
             - `external_id`

jcpitre:
`filter` and `sort` have been delegated to this issue: https://github.com/MobilityData/mobility-feed-api/issues/63

Issue #63, opened 2023-07-31
Design filtering options for REST API.


jcpitre:
### Describe the problem

In #49 we made some adjustments to the API. 
This was supposed to include filtering options but it was not done. 
I transferred the filtering part of the original issue here:

**feeds**

 string `filter`
      - filter options
          - contains string `provider`, e.g `"Kearns"`

**feeds/gtfs**

 filter options (should be able to filter by multiple values for each)
- contains string `provider`, e.g `"Kearns"`
- exact string `country_code`e.g `"FR"`
- contains string `subdivision_name` e.g `"Queensland"`
- contains string `municipality` e.g `"New York City"`
 - datetime `download_date`
      - greater than 1 date
      - lesser than 1 date
      - between 2 dates

**feeds/gtfs_rt**

 string `filter`
      - filter options
          - contains string `provider`, e.g `"Kearns"`
          - enum `entity_types`: options `vp`, `tu`, `sa`

**feeds/gtfs/{id}/datasets**

- filter options
    datetime `download_date`
   - greater than 1 date
   - lesser than 1 date
   - between 2 dates

### Proposed solution

Here are some suggestions:

- Example of filtering in REST api:
      - https://agileappscloud.info/aawiki/REST_API:Filter_Expressions_in_REST_APIs

- An in-house developed syntax for MVP (minimum viable product):
   - e.g. /feeds/gtfs/?filter=provider~mta&status=active
      - Use and ! (not), & (and) and | (or). These characters will need to be escaped since they are reserved for URIs.
      - Priority is !, &, then |.
      - Tests are:
         - =: exact value (all types)
         - <, >, <=, >=: 
            - Used for numerical fields and dates.
            - Error for strings (for MVP’s sake)
            ~: For strings only. field contains the string 

   - We need to make sure we can beef this up later (after MVP) and not introduce contradictions. We could later
      - Add parenthesis
      - Add regex
      - etc

- Github syntax:
   - They have a separate endpoint for search [link](https://docs.github.com/en/free-pro-team@latest/rest/search/search?apiVersion=2022-11-28#search-issues-and-pull-requests)
      - e.g. https://api.github.com/search/issues?q=Q
   - Elaborate search syntax: [link](https://docs.github.com/en/search-github/getting-started-with-searching-on-github/understanding-the-search-syntax)

- oData
  - [link](https://learn.microsoft.com/en-us/odata/concepts/queryoptions-overview)
[...]

PR #79, opened 2023-08-13
feat: implement feeds endpoint

aronza:
**Summary:**

Implements:

- https://github.com/MobilityData/mobility-feed-api/issues/32 
- https://github.com/MobilityData/mobility-feed-api/issues/35
- https://github.com/MobilityData/mobility-feed-api/issues/51
- fixes the broken tests by introducing environment variables during the tests and fix endpoints

**Expected behavior:** 

Screenshots from Postman requests against local API:

Screenshot 2023-08-13 at 3 23 47 PM

Screenshot 2023-08-13 at 3 23 00 PM
a7">

Screenshot 2023-08-13 at 3 24 07 PM

### DIFF
--- a/api/src/database/database.py
+++ b/api/src/database/database.py
@@ -89,13 +89,16 @@ class Database:
             self.logger.error(f"Session closing failed with exception: \n {e}")
         return self.is_connected()
 
-    def select(self, model: Type[Base], conditions: list = None, attributes: list = None, update_session: bool = True):
+    def select(self, model: Type[Base], conditions: list = None, attributes: list = None, update_session: bool = True,
+               limit: int = None, offset: int = None):
         """
         Executes a query on the database
         :param model: the sqlalchemy model to query
         :param conditions: list of conditions (filters for the query)
         :param attributes: list of model's attribute names that you want to fetch. If not given, fetches all attributes.
         :param update_session: option to update session before running the query (defaults to True)
+        :param limit: the optional number of rows to limit the query with
+        :param offset: the optional number of rows to offset the query with
         :return: None if database is inaccessible, the results of the query otherwise
         """
         try:
@@ -107,6 +110,10 @@ class Database:
                     query = query.filter(condition)
             if attributes:
                 query = query.options(load_only(*attributes))
+            if limit is not None:
+                query = query.limit(limit)
+            if offset is not None:
+                query = query.offset(offset)
             return query.all()
         except Exception as e:
             self.logger.error(f'SELECT query failed with exception: \n{e}')
@@ -187,3 +194,4 @@ class Database:
             return False
 
 
+DB_ENGINE = Database()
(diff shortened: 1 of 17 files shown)


jcpitre:
There are also some query parameters to implement, namely `limit` and `offset`.
`filter` and `sort` have been delegated to https://github.com/MobilityData/mobility-feed-api/issues/63
I suggest either you mention clearly in this PR and in the issue that limit and offset are not included, and not close the issue after the PR is closed. or you create a new issue to cover these parameters.

api/src/feeds/impl/feeds_api_impl.py
```diff
@@ -12,6 +16,24 @@
 from feeds_gen.models.latest_dataset import LatestDataset
 from feeds_gen.models.source_info import SourceInfo
 
+from database.database import Database
+
+db = Database()
```
cka-y, line 21: Nice to have: we should probably set up a mechanism to have a single instance of the database (singleton pattern). Instantiating the database as a global variable might increase the risk of concurrency issues and creates a lack of control over initialization in case we need to make any changes to it.

api/src/feeds/impl/feeds_api_impl.py
```diff
@@ -12,6 +16,24 @@
...
+
+
+def map_feed(feed: Feed):
```
cka-y, line 24: Suggestion: this function could be added to a `utils` file/module if you think it's going to be used by other classes or added as a static method of the `FeedApiImpl` class otherwise to keep the coding style consistent.

api/src/database/database.py
```diff
@@ -67,6 +68,7 @@ def start_session(self):
                 if self.get_connection is not None:
                     self.engine = create_engine("postgresql+pg8000://", creator=self.get_connection)
                 else:
+                    print(self.SQLALCHEMY_DATABASE_URL)
```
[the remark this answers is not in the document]
aronza: Removed, I was using it to debug environment variable issues

api/src/feeds/impl/feeds_api_impl.py
aronza: moved to a static method in this class as this looks specific to this module

api/src/feeds/impl/feeds_api_impl.py
aronza: moved to a global variable in database module

aronza:
> There are also some query parameters to implement, namely `limit` and `offset`. `filter` and `sort` have been delegated to #63 I suggest either you mention clearly in this PR and in the issue that limit and offset are not included, and not close the issue after the PR is closed. or you create a new issue to cover these parameters.

Added `limit` and `offset` to this PR. `filter` and `sort` can be tracked separately as these apply to most endpoints

aronza:
@cka-y we saw that you added changes to provider field in `feat_15.sql`. The OpenAPI spec still shows `providers` as a array of strings on `BasicFeed`. I've changed it to `provider` as a string

jcpitre (approved):

PR #86, opened 2023-08-22
fix: Optimize feeds query

aronza:
**Summary:**

Currently the API runs a database query for every external foreign key it needs to fetch. This causes the `/v1/feeds` endpoint to return 1903 feed objects in 12.99 seconds (according to PostMan)

Screenshot 2023-08-21 at 9 40 18 PM

**Expected behavior:** 
With these changes the endpoint returns under 600 milliseconds for the same amount of feed objects.

Screenshot 2023-08-21 at 9 42 13 PM

### DIFF
--- a/api/src/database/database.py
+++ b/api/src/database/database.py
@@ -5,6 +5,7 @@ from typing import Type
 from google.cloud.sql.connector import Connector
 from sqlalchemy import create_engine, inspect
 from sqlalchemy.orm import Session, load_only
+from sqlalchemy.sql import Select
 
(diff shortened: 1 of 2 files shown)


cka-y (approved):
LGTM! Question: was the issue related to the `redirects`?

aronza:
> LGTM! Question: was the issue related to the `redirects`?

It was both `redirects` and `external_ids`.

jujiang526:
how is the performance after this change?

Issue #84, opened 2023-08-21
Renaming API endpoints to resolve conflicting subpaths


indraneel:
### Describe the problem

- `/feeds/{id}` and `/feeds/gtfs` conflict because `{id}` is a string, but so is `gtfs`
- there is a design goal to have `feeds/` be an endpoint that returns every type of feed, including future ones

### Proposed solution

- One proposed fix is what @aronza has done here: https://github.com/MobilityData/mobility-feed-api/compare/main...arda/fix-api-conflict 
- - this renames `feeds/gtfs/` to `gtfs-feeds/`, and so on for the remaining current sub-paths of `feeds/`
- - preserves `feeds/` as is   

### Alternatives you've considered

Something like `feeds/id/gtfs` and `feeds/{id}/gtfs-rt`

emmambd:
I'm in support of this solution, as was @jcpitre I believe before his vacation. cc @qcdyx and @cka-y in case there's any outstanding strong opinions on our side.

indraneel:
Assigning to @aronza to open a PR w/ above changes

PR #89, opened 2023-08-25
fix: Renaming API endpoints to resolve conflicting subpaths

aronza:
**Summary:**

Issue: https://github.com/MobilityData/mobility-feed-api/issues/84

**Expected behavior:** 

All endpoints are accessible now.

Also all unit tests pass now. I removed a test for an endpoint that no longer exists.

I did not enable unit tests in CI because they depend on existing data in docker database. We can enable it when we have implemented complete unit tests (@goorui is working on it)


qcdyx (approved):
LGTM

PR #144, opened 2023-10-24
feat: add CORS support

### LINKED ISSUE #137: CORS error triggers when sending a request to Feeds API via browser
### Describe the bug

A CORS error triggers when sending a request to Feeds API via a browser; see attachment. It looks like the cause of the error is the authentication header is missing, as the preflight request receives an HTTP 401 response.a 

### Steps/Code to Reproduce

- Open the swagger editor and paste in the OpenAPI schema
- Open the developer console and network tab on the browser
- Sign in to https://qa.mobilitydatabase.org and obtain an access token
- Use the `try out` feature to test qa environment 

### Expected Results

The Feeds API response with HTTP 200 and shows results(can be empty array depending on the filters applied)

### Actual Results

An HTTP 401 is shown in the preflight request.

### Screenshots

Screenshot 2023-10-20 at 9 18 40 AM

### Files used

_No response_

### Additional notes

This might be related to the fact that we are using a Standard Load Balancer. Standard LB doesn't support CORS when IDP security is in place, see https://cloud.google.com/load-balancing/docs/features#security.

davidgamez:
**Summary:**

This PR adds CORS support via Load balancer configuration. 
 - [x] Add Python fastApi support for CORS calls
 - [x] Update OpenApi schema with Bearer Token authentication. This required the creation of two OpenApi schemas, one for `external` clients and a second for generating the code that resides under Google IAP proxy.
 - [x] Change Load Balancer from Classic to External to allow CORS rules to be set. This is not backward compatible with the currently deployed Load Balancer; we need to **delete** the current Load Balancer before deployment.
 - [x] Reserve IP addresses to have them independent of the Load Balancer. This is to allow deleting future Load Balancers if needed without changing the DNS

Fixes #137 

**Expected behavior:** 

As described in #137 CORS should be supported. Tested successfully against development environment.
Screenshot 2023-10-24 at 3 07 40 PM


docs/DatabaseCatalogAPI_IAP.yaml
```diff
@@ -0,0 +1,792 @@
+openapi: 3.0.0
```
jcpitre, line 1: Having two files that are identical except for one line hurts. There's no way to create a copy and replace the line with a sed?

davidgamez: My only concern is how clear it is to have a script that modifies the schema on the fly. At the same time, it's indeed painful to have two files with 99.9 % of matching lines. I created an issue to address this after MVP, #146.

api/src/main.py
```diff
@@ -0,0 +1,51 @@
+#
```
jcpitre, line 1: openapi generator still creates a main.py that is similar. Is it still used?
You should consider preventing the creation of main.py. I think it can be done in in gen-config.yaml.

scripts/api-gen.sh
```diff
@@ -8,21 +8,12 @@
...
-## More info,
-##      https://github.com/OpenAPITools/openapi-generator/issues/13863
-##      https://github.com/OpenAPITools/openapi-generator/issues/13863
-GENERATOR_VERSION=$BETA_VERSION
+GENERATOR_VERSION=7.0.1
 
 # relative path
 SCRIPT_PATH="$(dirname -- "${BASH_SOURCE[0]}")"
 
-OPENAPI_SCHEMA=$SCRIPT_PATH/../docs/DatabaseCatalogAPI.yaml
```
jcpitre, line 25: Perfect spot to generate DatabaseCatalogAPI_IAP.yaml.
Just sayin

jcpitre (approved):

api/src/main.py
davidgamez: I'm looking at the CLI documentation and I don't see any option to prevent the generation of a single file, https://openapi-generator.tech/docs/generators/python-fastapi.

jcpitre: Never trust chatgpt. Never mind.

PR #150, opened 2023-11-02
Added redirect field (id and comment)

jcpitre:
**Summary:**
Closes https://github.com/MobilityData/mobility-feed-api/issues/149
Associated PR in mobility catalogs: https://github.com/MobilityData/mobility-database-catalogs/pull/309

- Added a redirect comment to the `redirectingid` table in the DB
- Modified populate script to take into account the `redirect.id` and `redirect.comment` columns in the csv.
- Modified the openapi specs to add a redirect id and comment to the schemas
- Modified the API code to return the data.

**Expected behavior:** 

With https://github.com/MobilityData/mobility-database-catalogs/pull/309 the csv should have redirect.id and redirect.comment columns

- The DB should now have a `redirect_comment` added to the `redirectingid` table.
- The populate script should be able to read the 2 new columns and fill the redirecting table properly.
- Calling the API for `feeds`, `gtfs_feeds` and `gtfs_rt_feeds` should now return a `redirects` array.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Run the unit tests with `./scripts/api-tests.sh` to make sure you didn't break anything
- [ ] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] Linked all relevant issues
- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)

PR #159, opened 2023-11-06
feat: Changed a description in the openapi file.

jcpitre:
**Summary:**

Changed according to comment in https://github.com/MobilityData/mobility-feed-api/pull/150


emmambd (approved):
@jcpitre Thanks for making this change! Looks great.

Issue #190, opened 2023-11-29
Sending invalid limit parameter results in error 500


jcpitre:
### Describe the bug

Making this call to the API (notice the `limit=-1` parameter):
```
curl -X 'GET' \
  'https://api-qa.mobilitydatabase.org/v1/gtfs_feeds?limit=-1' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer ...'
```
Results in a error 500, which usually means an error in the server that should be corrected by the server developer. 
In this case it's an error dependant on the parameters passed in the call, so It should be an error in the 4XX range.

FYI If the offset was negative:
```
curl -X 'GET' \
  'https://api-qa.mobilitydatabase.org/v1/gtfs_feeds?offset=-1' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer ...'
```
Then it currently returns an error 422 with this json:
```
{
  "detail": [
    {
      "loc": [
        "query",
        "offset"
      ],
      "msg": "ensure this value is greater than or equal to 0",
      "type": "value_error.number.not_ge",
      "ctx": {
        "limit_value": 0
      }
    }
  ]
}
```
which makes more sense.

### Steps/Code to Reproduce

See above

### Expected Results

See above

### Actual Results

See above

aronza:
Created this [PR](https://github.com/MobilityData/mobility-feed-api/pull/201). @jcpitre I see we have two Open API definitions now under docs folder. I can see that they are using different token schemas. Are we maintaining both currently or only the DatabaseCatalogAPI_IAP.yaml that is referenced from scripts/api-gen.sh?

jcpitre:
@aronza Both yaml files are currently needed and have to be updated.
I created an issue for combining them https://github.com/MobilityData/mobility-feed-api/issues/203

PR #201, opened 2023-12-26
fix: 500 when limit is -1

aronza:
**Summary:**

Fixes the 500 response from the API when limit parameter passed is negative. https://github.com/MobilityData/mobility-feed-api/issues/190

**Expected behavior:** 

Returns 422 with the below JSON body like offset parameter.

Screenshot 2023-12-26 at 2 02 26 PM


jcpitre (approved):
LGTM!

PR #218, opened 2024-01-09
chore: remove GCP credentials from build-test GH action

davidgamez:
**Summary:**

The GCP pub-sub global variable was loaded before mocks were applied. This enforced GCP credentials set and blocks external contributors from running tests on the local environment and forks. This PR remove the GCP credentials dependency by making the pub-sub client scoped to the main python function.

**Expected behavior:** 

GCP credentials dependency is removed from local environment and GH forks.

**Testing tips:**

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -27,15 +27,6 @@ jobs:
         with:
           python-version: ${{ env.python_version }}
 
-      - name: Authenticate to Google Cloud
-        id: gcloud_auth
-        uses: google-github-actions/auth@v1
-        with:
-          credentials_json: ${{ secrets.DEV_GCP_MOBILITY_FEEDS_SA_KEY }}
-
-      - name: GCloud Setup
-        uses: google-github-actions/setup-gcloud@v1
-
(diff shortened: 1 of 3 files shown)


cka-y (commented):
Question: Does the python functions tests need to be a part of the `build-test` of the api? if so shouldn't the typescript functions be part of the testing as well? 
Suggestion: I think an easy fix would be to have a `build-test` workflow specific to the python functions and one for the typescript functions that would only get triggered on PR if a file within these folders has been modified.

davidgamez:
> Question: Does the python functions tests need to be a part of the `build-test` of the api? if so shouldn't the typescript functions be part of the testing as well? Suggestion: I think an easy fix would be to have a `build-test` workflow specific to the python functions and one for the typescript functions that would only get triggered on PR if a file within these folders has been modified.

The same question applies to the API code section. We definitely need to test the Python functions on PRs. We can optimize the testing execution by filtering the folder, taking also in consideration changes to the DB. I would leave that optimization outside this PR.

cka-y (approved):

PR #231, opened 2024-01-12
fix:217 make all labels snake case

aronza:
**Summary:**

Make all labels in OpenAPI definition snake case

**Expected behavior:** 

No API change. Swagger document should show snake case

### DIFF
--- a/docs/DatabaseCatalogAPI.yaml
+++ b/docs/DatabaseCatalogAPI.yaml
@@ -93,9 +93,9 @@ paths:
         - $ref: "#/components/parameters/subdivision_name"
         - $ref: "#/components/parameters/municipality"
         - $ref: "#/components/parameters/sort"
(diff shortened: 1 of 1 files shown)


jcpitre:
Also to do:
- remove `filter` parameter
- remove `sort` parameter in the yaml and python code. The python code does nothing with it.

Also in general some labels or names are in camelCase and others in snake_case. I would put everything snake_case. e.g. latestQueryParam -> latest_query_param. Even if it's not _customer facing_ why not have something uniform.

indraneel:
@jcpitre would it be ok if we moved this PR through once the filter/sort is removed and tackle the camel_case feedback in this issue #238?

jcpitre:
@indraneel sure

jcpitre (approved):

Issue #53, opened 2023-07-19
Update info links on Swagger


emmambd:
- Update terms of service
- Update email address
- Update MobilityData license to point to https://www.apache.org/licenses/LICENSE-2.0

PR #246, opened 2024-01-22
doc: update docs (info links, examples, etc)

royliTS:
**Summary:**

For [issue 53](https://github.com/MobilityData/mobility-feed-api/issues/53): Update docs and remove sort in API

**Expected behavior:** 

- Contact email, license link, and production server have been updated
- In examples, USA has been replaced with US and an offset of 0 is used
- In docs, provider filter has been specified to use the LIKE operator
- Sort field has been removed from all API endpoints
- Order by options without "+" or "-" have been removed
Screenshot 2024-01-22 at 2 29 40 PM

**Testing tips:**

Verify everything is as expected in the UI


docs/DatabaseCatalogAPI.yaml
```diff
@@ -831,7 +818,7 @@ components:
         type: integer
         minimum: 0
         default: 0
-        example: 10
+        example: 0
```
jcpitre, line 821: The error message for when I enter. negtive number is still:
`Value must be greater than 0`
Is there any way it could say something like:
`Value must be equal or greater than 0`?

royliTS: Spent a bit of time looking into it and seems like the default error message is hardcoded in the Swagger javascript somewhere. There is an `excludeMininum` option for the numbers type, but its value doesn't affect the message. 

It would take some effort to change this default error message (probably easier to not use the built-in validation and validate on the server side instead, but this doesn't sound ideal). Given that the default value and the example value are both 0 (and submitting a request with offset=0 doesn't produce a bad response), maybe users can infer that 0 is accepted. If we want, we can start a separate issue to look into updating the default error messages?

jcpitre: Yes, a separate issue is good. It's really a minor problem.

jcpitre (approved):
LGTM!

PR #259, opened 2024-02-05
feat: added integration tests

cka-y:
**Summary:**

copilot: summary

**Expected behavior:** 

**Testing tips:**


integration-tests/src/endpoints/feeds.py
```diff
@@ -0,0 +1,163 @@
...
+        """Test feed retrieval with limit"""
+        response = self.get_response("v1/feeds", {"limit": 10})
+        assert (
+            response.status_code == 200
+        ), f"Expected 200 status code, got {response.status_code}."
+        feeds = response.json()
+        assert isinstance(feeds, list), "Response should be a list."
+        assert len(feeds) == 10, f"Expected 10 feeds, got {len(feeds)}."
+
+    def test_feeds_offset(self):
```
jcpitre, line 43: Having disjoint lists returned only partially test this functionality. Have you considered something like having an offset of 1 and checking that only the first item of the first list is missing in the second list?
Or even simpler for the first retrieval set a limit of 2 and the second retrieval set an offset of 1 and a limit of 1. The sole item from the second list should be the same as the second item of the first list.

integration-tests/src/endpoints/feeds.py
```diff
@@ -0,0 +1,163 @@
...
+        prev_feed_status = feeds[0]["status"]
+        for feed in feeds[1:]:
+            current_feed_status = feed["status"]
+            assert current_feed_status <= prev_feed_status, (
+                f"Expected feed status to be in descending order, but found '{prev_feed_status}' followed by "
+                f"'{current_feed_status}'."
+            )
+            prev_feed_status = current_feed_status
+
+    def test_feeds_status_sorting_ascending(self):
```
jcpitre, line 99: `sort` is slated to be removed.
It's replaced by `order_by` that can order by external_id and country_code.

jcpitre (approved):
Impressive!

integration-tests/src/endpoints/feeds.py
cka-y: Good point! I added `test_feeds_pagination_simple_offset_limit` and `test_feeds_pagination_incremental_offset`.

integration-tests/src/endpoints/feeds.py
jcpitre: I would remove `sort` and add tests for `order_by`

integration-tests/src/endpoints/feeds.py
jcpitre: Great. For me it was actually an OR, ie do one or the other, but the more the better, right?

jcpitre (approved):

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,58 @@
...
+
+    - name: Install dependencies
+      working-directory: integration-tests
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+
+    - name: Run Integration Tests
+      run: |
+        ./scripts/integration-tests.sh -u ${{ inputs.API_URL }} -t ${{ secrets.REFRESH_TOKEN }} -f $FILE_PATH
```
jcpitre, line 44: I hear it's not good practice top put a secret as an argument of the command line since it can then be visible with `ps`

scripts/integration-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+source "$VENV_PATH"/bin/activate
+pip install -r "$PARENT_DIR"/integration-tests/requirements.txt &> /dev/null
+
+export PYTHONPATH="${PARENT_DIR}:${PARENT_DIR}/integration-tests/src:${PARENT_DIR}/api/src"
+
+# Pass include_classes if it's not empty
+if [ -z "${INCLUDE_CLASSES}" ]; then
+  (cd "$PARENT_DIR"/integration-tests/src && python "$PARENT_DIR"/integration-tests/src/main.py --file_path "${FILE_PATH}" --access_token "$ACCESS_TOKEN" --url "$URL")
+else
+  (cd "$PARENT_DIR"/integration-tests/src && python "$PARENT_DIR"/integration-tests/src/main.py --file_path "${FILE_PATH}" --access_token "$ACCESS_TOKEN" --url "$URL" --include_classes "${INCLUDE_CLASSES}")
```
jcpitre, line 62: Same comment with this token , although it's short(er) lived.

.github/workflows/integration-tests.yml
cka-y: Changed to env variable and improved error logs!

scripts/integration-tests.sh
cka-y: Changed to env variable!

integration-tests/src/endpoints/datasets.py
```diff
@@ -0,0 +1,140 @@
...
+                else FeedError.API_ERROR.value
+            )
+            raise Exception(f"{error_code}: {error_detail}")
+
+    @staticmethod
+    def _validate_dataset(datasets, status_code):
+        if status_code != 200 or not datasets:
+            error_code = FeedError.NO_DATASET.name
+            error_detail = FeedError.NO_DATASET.value
+            raise Exception(f"{error_code}: {error_detail}")
```
davidgamez, line 100: I'm not sure if no having a dataset is an error. It's a valid use case.

cka-y: The exception is caught and doesn't make the script fail! It's just to create the report and express that this feed does not have a dataset

davidgamez: I see. Is this also the case with the feed validation?

cka-y: Updated: the code throws an exception after all test have ran if any failed! (datasets validation excluded)

Issue #66, opened 2023-08-01
Add latest dataset metadata to feed response


emmambd:
### Describe the problem

- If I want to just check a feed and re-pull the `hosted_URL` each time there's a new dataset or the URL is changed, right now I have no way of seeing when the hosted_URL has been updated. 

### Proposed solution

- Add `downloaded_at`(~~`updated_at`~~)( field in latest_dataset object in all feeds endpoints to track when a log was made (that way I know the feed's been changed and I should refresh my data).
- Add dataset `hash` to `latest_dataset` object in feed response

davidgamez:
Tasks:
- [ ] Update OpenAPI schema.
- [ ] Refactor feed's endpoint.

jcpitre:
@emmambd I'm having second thoughts about `downloaded_at`. The __at_ suffix is OK if we specify a time (downloaded at 3PM for example) but for a date doesn't `downloaded_on` be more appropriate? (downloaded on the 4th of July).

Nitpicking maybe, but I would hate to be stuck forever with a syntax that doesn't feel quite right.

emmambd:
@jcpitre Based on Arda's PR: https://github.com/MobilityData/mobility-feed-api/pull/262 it's a timestamp, so it'll have both a date and time. Does it still seem off to you?

jcpitre:
> [@jcpitre](https://github.com/jcpitre) Based on Arda's PR: [#262](https://github.com/MobilityData/mobility-feed-api/pull/262) it's a timestamp, so it'll have both a date and time. Does it still seem off to you?

Yes. I guess there is no perfect solution. `download_timestamp`? Kind of a mouthful.

emmambd:
Decision: leave downloaded_at as is.

PR #262, opened 2024-02-06
add latest dataset metadata to feed response

aronza:
**Summary:**

Issue: https://github.com/MobilityData/mobility-feed-api/issues/66

**Expected behavior:** 

New dataset columns should be visible from boh `/gtfs_feeds` and `/gtfs_feeds/{id}` endpoints

Screenshot 2024-02-06 at 2 52 49 PM

Screenshot 2024-02-06 at 2 54 36 PM

Updated the database and all referenced places based on the request at https://github.com/MobilityData/mobility-feed-api/issues/66#issuecomment-1733928274

Though I don't see any mention of GtfsDataset's in populate_db.py


jcpitre (approved):

aronza:
> @aronza The error with the tests is also in other branches. I think it was introduced into main between yesterday and now

Yep, I've tried debugging it but only found the issue with environment variables not being set correctly. It wasn't enough to fix the ImportError though

Issue #239, opened 2024-01-17
Modify OpenAPI Schema so the "live" version of the API docs doesn't include the validation report metadata


emmambd:
### Describe the problem

We need some way to separate out "draft" schema changes from those that represent the V1 of the API that users can rely on. 

### Proposed solution

New version of the Open API schema that has a label of [DRAFT v1.1]?

jcpitre:
We decided that we would keep the implementation with the ValidationReport section and just document it as not supported for now.

PR #289, opened 2024-02-13
feat: Made API examples work out of the box

### LINKED ISSUE #284: Swagger documentation changes
- Update gtfs_feeds example bounding box input to -78.58,-87.29 instead of -78.58,-87-29 (should check to see if this actually returns a feed example)
- Check all endpoint example “try it out” buttons  to make sure that they each return a full response
- Update terms of service link to mobilitydatabase.org/terms-and-conditions

jcpitre:
**Summary:**

Closes #284 

**Expected behavior:** 

Using swagger, doing a `Try it out` on any of the API endpoints shouild give some result (ie no error or empty list)

### DIFF
--- a/docs/DatabaseCatalogAPI.yaml
+++ b/docs/DatabaseCatalogAPI.yaml
@@ -18,7 +18,7 @@ servers:
   - url: https://api-qa.mobilitydatabase.org/
     description: Pre-prod environment
   - url: https://api-dev.mobilitydatabase.org/
-    description: Development environment    
+    description: Development environment
   - url: http://localhost:8080/
     description: Local development environment
 
@@ -114,8 +114,8 @@ paths:
       parameters:
         - $ref: "#/components/parameters/limitQueryParam"
         - $ref: "#/components/parameters/offset"
-        - $ref: "#/components/parameters/provider"
-        - $ref: "#/components/parameters/producer_url"
+        - $ref: "#/components/parameters/provider_rt"
+        - $ref: "#/components/parameters/producer_url_rt"
         - $ref: "#/components/parameters/entity_types"
       security:
         - Authentication: []
@@ -148,7 +148,7 @@ paths:
 
   /v1/gtfs_rt_feeds/{id}:
     parameters:
-      - $ref: "#/components/parameters/feedIdPathParam"
+      - $ref: "#/components/parameters/feedIdPathParam_rt"
     get:
       description: Get the specified GTFS Realtime feed from the Mobility Database.
       tags:
@@ -620,14 +620,6 @@ components:
 #      example: gtfs
   parameters:
 
-    filter:
-      name: filter
-      in: query
-      description: A filter to apply to the returned data. Exact syntax to be designed
-      required: False
-      schema:
-        type: string
-        example: status=active
     status:
       name: status
       in: query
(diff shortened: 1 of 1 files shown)


docs/DatabaseCatalogAPI.yaml
```diff
@@ -643,29 +635,51 @@ components:
     provider:
       name: provider
       in: query
-      description: Filter feeds by provider name using the LIKE operation. Phrase is matched case insensitively to actual provider names.
+      description: List only feeds with the specified value. Can be a partial match. Case insensitive.
       required: false
       schema:
         type: string
-        example: London Transit
+        example: Los Angeles Department of Transportation
```
emmambd, line 642: Would be nice to return an example for /feeds that shows multiple feeds rather than just one. How about Intercity Transit, with the producer_url of https://its.rideralerts.com?

emmambd (changes requested):
Looks good! 2 minor changes. Also, we should change the default example that appears in each endpoint before you click the "Try it out" button - it isn't accurate. Could just be prepopulated with each Try it out example. 
Screenshot 2024-02-13 at 5 17 07 PM

emmambd:
Another thought: @jcpitre : Have we added anywhere that the validation report section is not supported in the Swagger documentation?

docs/DatabaseCatalogAPI.yaml
jcpitre: If I just filter with `Intercity Transit` it does return 2 feeds. But if I add the producer_url then it returns only one feed.
We want examples for each parameter, right? 
So that means we would need a query that returns 2 or more feeds, but that would have the same provider, producer_url, country_code, subdivision_name, municipality, latitudes, longitudes. 
Any parameter that we would want different would mean no example for that parameter.
Does this make sense?

docs/DatabaseCatalogAPI.yaml
```diff
@@ -620,14 +620,6 @@ components:
...
-      name: filter
-      in: query
-      description: A filter to apply to the returned data. Exact syntax to be designed
-      required: False
-      schema:
-        type: string
-        example: status=active
     status:
       name: status
       in: query
```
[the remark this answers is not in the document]
jcpitre: Apparently openapi accepts markdown.
Added.

jcpitre:
> Looks good! 2 minor changes. Also, we should change the default example that appears in each endpoint before you click the "Try it out" button - it isn't accurate. Could just be prepopulated with each Try it out example. 

Yes. It's a bit more work than it looks because the data for the default example is distributed over all the schemas. I can't just copy the json as is.

docs/DatabaseCatalogAPI.yaml
```diff
@@ -643,29 +635,51 @@ components:
...
-      description: Filter feeds by provider name using the LIKE operation. Phrase is matched case insensitively to actual provider names.
+      description: List only feeds with the specified value. Can be a partial match. Case insensitive.
       required: false
       schema:
         type: string
-        example: London Transit
+        example: Los Angeles Department of Transportation
+
+    # A provider specific to gtfs realtime simply to have a different example
```
davidgamez, line 644: I think we should not add  _"extra"_ parameters just for the purpose of having a different example in the SwaggerUI. SwaggerUI users should not be the main consumers of the schema.

docs/DatabaseCatalogAPI.yaml
emmambd: True, that does make sense. We can leave the example as is then.

jcpitre:
> Another thought: @jcpitre : Have we added anywhere that the validation report section is not supported in the Swagger documentation to close out #239 ?

Description for validation report is now: `Validation report - Not yet supported. Can change in the future.`

emmambd (approved):
@jcpitre LGTM!

emmambd:
@jcpitre One thing, description of the API in the header is old: "API as required in the Proposed Version 1 from the Product Requirement Document for the Mobility Database". 

Title should be "Mobility Database Catalog"
Description should link to the website: https://database.mobilitydata.org/

emmambd (changes requested):
Two things: description of the API in the header is old: "API as required in the Proposed Version 1 from the Product Requirement Document for the Mobility Database".

Title should be "Mobility Database Catalog"
Description should link to the website: https://mobilitydatabase.org/

I also don't see the terms of service link update.

PR #375, opened 2024-04-04
fix: workflow creation permission

cka-y:
Hotfix for the failing deployment in QA due to the service account not having permission to create a workflow in GCP.


davidgamez (approved):
LGTM!

PR #382, opened 2024-04-09
feat: search endpoint implementation

### LINKED ISSUE #371: Implement search endpoint
Based on work in #363

davidgamez:
**Summary:**

Changes
- Add a materialized view for indexing feeds. Added all returned fields to the view to reduce DB calls.
- Implemented search endpoint querying the view with pattern '{query}:*', allowing incomplete lexemes in the queries.
- Refresh the materialized view after inserting data.

**Performance considerations**
We included all fields in the view to avoid multiple DB calls per HTTP request, which impacts the view's size. However, the current size is still small: ~1.6MB. We should monitor this size as the DB grows in the number of feeds.
DEV test was performed with the following results:
|    |  Time  |  Response Size  |  
|---|---|---|
|  limit=100, search_query=us | <150ms   |  90kb
|  limit=200, search_query=us  | <250ms   |  180kb
|  no limit, search_query=us  | ~1s   |  896kb
|  no limit, no search_query | ~1.6s   |  1.6MB

**Alternative solution considered**
We considered having a minimal footprint in the view to support the indexing and subsequence search action. Then, the Python code would call the DB multiple times to populate all remaining fields. Reducing the fields returned by the search endpoint is a possible optimization. 

Closes #371

**Expected behavior:** 

**Testing tips:**
Test locally the new search endpoint
- Update your local DB schema:
`docker-compose --env-file ./config/.env.local  up -d --force-recreate`
- Populate your DB by downloading the [csv catalog](https://bit.ly/catalogs-csv) and running the populate script
`./scripts/populate-db.sh {full-path of the catalog}`
- Start the Python application by using `./scripts/api-start.sh`
- Execute and play with different queries and parameters, example:
`curl --location 'http://localhost:8080/v1/search?search_query=mta&offset=0&limit=5' \
--header 'Accept: application/json'
`

### DIFF
--- a/api/src/feeds/impl/search_api_impl.py
+++ b/api/src/feeds/impl/search_api_impl.py
@@ -17,11 +17,20 @@ class SearchApiImpl(BaseSearchApi):
 
     @staticmethod
     def get_parsed_search_tsquery(search_query: str) -> str:
+        """
+        Parse the search query to be used in the database query.
+        The resulting query will be in the form of `search_query:*` if the search query is not empty.
+        Spaces are trimmed from the search query.
+        """
         parsed_query = f"{search_query.strip()}:*" if search_query and len(search_query.strip()) > 0 else ""
         return func.plainto_tsquery("english", parsed_query)
 
     @staticmethod
     def add_search_query_filters(query, search_query, data_type, feed_id, status) -> Query:
+        """
+        Add filters to the search query.
+        Filter values are trimmed and converted to lowercase.
+        """
         if feed_id:
             query = query.where(t_feedsearch.c.feed_stable_id == feed_id.strip().lower())
         if data_type:
--- a/api/src/feeds/impl/search_api_impl.py
+++ b/api/src/feeds/impl/search_api_impl.py
@@ -7,13 +7,18 @@ from feeds.impl.models.search_feeds200_response_results_inner_impl import Search
 from feeds_gen.apis.search_api_base import BaseSearchApi
 from feeds_gen.models.search_feeds200_response import SearchFeeds200Response
 
-feed_search_columns = [column for column in t_feedsearch.columns if column.name != 'document']
+feed_search_columns = [column for column in t_feedsearch.columns if column.name != "document"]
 
 
 class SearchApiImpl(BaseSearchApi):
     """
     This class represents the implementation of the `/search` endpoints.
     """
+
+    @staticmethod
+    def get_parsed_search_tsquery(search_query: str) -> str:
+        parsed_query = f"{search_query.strip()}:*" if search_query and len(search_query.strip()) > 0 else ""
+        return func.plainto_tsquery("english", parsed_query)
 
     @staticmethod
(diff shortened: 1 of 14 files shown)


api/src/feeds/impl/search_api_impl.py
```diff
@@ -0,0 +1,82 @@
...
+            search_query: str,
+    ) -> SearchFeeds200Response:
+        """Search feeds using full-text search on feed, location and provider&#39;s information."""
+        query = self.create_search_query(status, feed_id, data_type, search_query)
+        feed_rows = Database().select(
+            query=query,
+            limit=limit,
+            offset=offset,
+        )
+        feed_total_count = Database().select(
```
Alessandro100, line 75: wouldn't it be better to have 1 database class instance and reusing it vs 2 `Database()`? Also whats the protocol to initiate a session directly in `database.by` instead of using `helpers/database.py`

davidgamez: The `helpers/database.py` is a different implementation of the database connection. It belongs to the Python functions and is not part of the API source., even though it is in the same repository. In the future, we should try to unify the API with the Python functions to make the development process less confusing.

functions-python/test_utils/database_utils.py
```diff
@@ -56,17 +58,30 @@ def get_testing_session() -> Session:
 
 
 def clean_testing_db():
-    """Truncates all table in the test db."""
+    """Deletes all rows from all tables in the test db, excluding those in excluded_tables."""
```
davidgamez, line 61: Using DELETE instead of TRUNCATE to avoid database locks on failing tests.

api/tests/test_search_api.py
```diff
@@ -0,0 +1,310 @@
...
+    )
+
+    # Assert the status code of the HTTP response
+    assert response.status_code == 200
+
+    # Parse the response body into a Python object
+    response_body = SearchFeeds200Response.parse_obj(response.json())
+
+    assert response_body.total == len(TEST_GTFS_FEED_STABLE_IDS)
+    assert len(response_body.results) == len(TEST_GTFS_FEED_STABLE_IDS)
```
Alessandro100, line 44: what is the reasoning behind returning the property `total` if it can be acquired through `results.length`

davidgamez: The _total_ property contains the total of the _query_ without considering _limit_ and _offset_(pagination). The total is the same in this specific test with or without pagination.

Alessandro100 (approved):
Testing instructions worked well, LGTM ✅

jcpitre:
@davidgamez I seems the | (or) and ! (not) don't work like I would expect them to.
The not had no effect e.g. `!Queensland` returned the same as `Queensland`.
The or returned an empty array e.g. `Liguria | Lombardia`, even if searching for `Liguria` returned a list so did `Lombardia`

There might be something I am doing wrong, so I am adding this comment instead of opening a new issue.

PR #410, opened 2024-04-30
feat: Implement script to update existing datasets with validator reports

### LINKED ISSUE #395: Implement script to update existing datasets with validator reports
Issue #229 covers the task of updating the database with validator reports when a new dataset is added. However, the current implementation does not account for datasets that are already stored in the database and lack associated validation reports. A one-time script is needed to handle this situation.

**Tasks:**
- [x] Create a script to identify all datasets that do not have an associated validator report.
- [x] Trigger the `gtfs_validator_execution` workflow for each of these datasets, ensuring that the correct data is provided.

cka-y:
**Summary:**
This PR addresses issue #395 by introducing a new cloud function designed to update validation reports for all `latest` datasets in a batch process.

**Details:**
This enhancement is triggered (e.g., upon a GTFS Validator release) to perform batch validations. It retrieves all datasets tagged as `latest` and checks their corresponding validation reports stored in GCP against the most recent validation version available via the web validator. For any dataset lacking the latest report version, the function updates the associated GCP `Blob` metadata—including retry count and validation version. This metadata aids in managing update attempts, ensuring that datasets that are repeatedly invalid do not continue to retry indefinitely. Following metadata updates, the corresponding GCP workflow that downloads the report to GCP and updates the database is automatically initiated.

**Expected Behavior:**
Upon execution, the function will:
- Retrieve the expected latest version using the web validator.
- Update metadata for any `latest` datasets with outdated reports.
- This update will automatically trigger the necessary workflows to refresh the GCP data storage and database.

**Testing Instructions:**
1. Delete existing reports in GCP to simulate the need for updates.
2. Trigger the function using the following `curl` command:
   ```
   curl -m 3610 -X POST https://northamerica-northeast1-mobility-feeds-dev.cloudfunctions.net/update-validation-report \
   -H "Content-Type: application/json" 
   ```
3. Monitor the logs for any unusual activity.
4. Verify that the workflow was triggered as expected by visiting the [workflow page](https://console.cloud.google.com/workflows/workflow/northamerica-northeast1/gtfs_validator_execution/executions?project=mobility-feeds-dev).

### DIFF
--- a/functions-python/update_validation_report/src/main.py
+++ b/functions-python/update_validation_report/src/main.py
@@ -173,9 +173,7 @@ def get_dataset_blobs_for_validation(
         system_errors_blob = bucket.blob(
             f"{feed_id}/{dataset_id}/system_errors_{validator_version}.json"
         )
-        dataset_blob = bucket.blob(
-            f"{feed_id}/{dataset_id}/{dataset_id}.zip"
-        )
+        dataset_blob = bucket.blob(f"{feed_id}/{dataset_id}/{dataset_id}.zip")
         dataset_blob_exists = dataset_blob.exists()
         if not dataset_blob_exists:
             logging.warning(f"Dataset blob not found for {feed_id}/{dataset_id}")
--- a/functions-python/update_validation_report/src/main.py
+++ b/functions-python/update_validation_report/src/main.py
@@ -174,13 +174,15 @@ def get_dataset_blobs_for_validation(
             f"{feed_id}/{dataset_id}/system_errors_{validator_version}.json"
         )
         dataset_blob = bucket.blob(
-            f"{feed_id}/{dataset_id}/dataset_{validator_version}.json"
+            f"{feed_id}/{dataset_id}/{dataset_id}.zip"
         )
         dataset_blob_exists = dataset_blob.exists()
         if not dataset_blob_exists:
             logging.warning(f"Dataset blob not found for {feed_id}/{dataset_id}")
+        # The system errors blob is used to determine if the validation report needs to be updated
+        # since it's the only report that is always generated even if there are errors during validation
         if not system_errors_blob.exists() and dataset_blob_exists:
-            report_update_needed.append(system_errors_blob)
+            report_update_needed.append(dataset_blob)
             logging.info(
                 f"System errors blob not found for {feed_id}/{dataset_id} -- Adding to update list"
             )
--- a/functions-python/update_validation_report/src/main.py
+++ b/functions-python/update_validation_report/src/main.py
@@ -202,7 +202,9 @@ def execute_workflows(latest_datasets):
             input_data = {
                 "data": {
                     "protoPayload": {
-                        "resourceName": f"projects/_/buckets/{bucket_name}/objects/{feed_id}/{dataset_id}/{dataset_id}.zip"
+                        "resourceName": "projects/_/"
+                        f"buckets/{bucket_name}/"
+                        f"objects/{feed_id}/{dataset_id}/{dataset_id}.zip"
                     },
                     "resource": {
                         "labels": {"location": location, "project_id": project_id},
--- a/functions-python/update_validation_report/src/main.py
+++ b/functions-python/update_validation_report/src/main.py
@@ -197,6 +197,7 @@ def execute_workflows(latest_datasets):
     batch_size = int(os.getenv("BATCH_SIZE", 5))
     sleep_time = int(os.getenv("SLEEP_TIME", 5))
     count = 0
+    logging.info(f"Executing workflow for {len(latest_datasets)} datasets")
     for feed_id, dataset_id in latest_datasets:
         try:
             input_data = {
@@ -219,6 +220,7 @@ def execute_workflows(latest_datasets):
                 f"Error while executing workflow for {feed_id}/{dataset_id}: {e}"
             )
         count += 1
+        logging.info(f"Triggered workflow execution for {count} datasets")
         if count % batch_size == 0:
             logging.info(
                 f"Sleeping for {sleep_time} seconds before next batch to avoid rate limiting.."
--- a/functions-python/update_validation_report/src/main.py
+++ b/functions-python/update_validation_report/src/main.py
@@ -21,15 +21,23 @@ from typing import List
 import functions_framework
 import requests
 import sqlalchemy.orm
+import json
+from sqlalchemy import or_
 from google.cloud import storage
 from sqlalchemy.engine import Row
 from sqlalchemy.engine.interfaces import Any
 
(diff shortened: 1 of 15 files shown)


functions-python/update_validation_report/src/main.py
```diff
@@ -0,0 +1,189 @@
...
+    :return: The list of updated datasets stable ids
+    """
+    max_retry = os.getenv("MAX_RETRY", None)
+    batch_size = int(os.getenv("BATCH_SIZE", 5))
+    sleep_time = int(os.getenv("SLEEP_TIME", 5))
+    logging.info(f"Max retry: {max_retry}, Batch size: {batch_size}")
+
+    batch_count = 0
+    updated_datasets = []
+    for blob in dataset_blobs:
```
davidgamez, line 101: [suggestion]: The loop does not handle exceptions. I suggest wrapping a single dataset's processing within try/catch and ignoring the errors. This way, we can continue processing even if one or a few operations fail.

functions-python/update_validation_report/src/main.py
```diff
@@ -0,0 +1,187 @@
...
+    validator_version = get_validator_version()
+
+    bucket_name = f"mobilitydata-datasets-{env}"
+    logging.info(f"Accessing bucket {bucket_name}")
+
+    session = start_db_session(os.getenv("FEEDS_DATABASE_URL"), echo=False)
+    latest_datasets = get_latest_datasets(session)
+    logging.info(f"Retrieved {len(latest_datasets)} latest datasets.")
+
+    dataset_blobs = get_dataset_blobs_for_validation(
```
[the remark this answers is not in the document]
jcpitre: The blob themselves should be relatively small, no?

functions-python/update_validation_report/src/main.py
```diff
@@ -0,0 +1,187 @@
...
+    sleep_time = int(os.getenv("SLEEP_TIME", 5))
+    logging.info(f"Max retry: {max_retry}, Batch size: {batch_size}")
+
+    batch_count = 0
+    updated_datasets = []
+    for blob in dataset_blobs:
+        # Update metadata to trigger the validation report processor
+        metadata = blob.metadata if blob.metadata is not None else {}
+        retry = 0
+        if "retry" in metadata:
```
jcpitre, line 105: You should consider giving this a slightly more explicit name, like `validation_retry`.
Somebody looking at the metadata of the file in the bucket might not know what `retry` mean. `validation_retry` gives just a bit more information.

functions-python/update_validation_report/src/main.py
```diff
@@ -0,0 +1,187 @@
...
+from sqlalchemy.engine.interfaces import Any
+
+from database_gen.sqlacodegen_models import Gtfsdataset, Gtfsfeed
+from helpers.database import start_db_session
+from helpers.logger import Logger
+
+logging.basicConfig(level=logging.INFO)
+
+
+@functions_framework.http
```
jcpitre, line 35: I can't see who calls this endpoint.

cka-y: no process currently calls this endpoint -- it has to be manually called or part of the release process of the GTFS validator after the web validator has been updated

functions-python/update_validation_report/src/main.py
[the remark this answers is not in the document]
cka-y: @davidgamez The blob content itself isn't directly loaded; we only maintain a reference to each blob within the list. Considering we will have up to 2000 blobs (corresponding to our current count of 2000 feeds), the memory footprint should remain manageable.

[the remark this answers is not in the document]
davidgamez: I understand that the blob doesn't contain the data if it's not read. In this case, can we reduce the memory of the instance? 2.5GB might be too large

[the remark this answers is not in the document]
cka-y: Agreed! Updated it to 256MiB which is similar to the `batch-datasets` http cloud function. Both of these function should have a similar memory footprint, but i'll keep an eye on it.

functions-python/update_validation_report/src/main.py
cka-y: done ✅  Added a `try/catch` clause for each loop.

functions-python/update_validation_report/src/main.py
cka-y: Resolving this comment since the trigger is now done directly through the function (not with metadata update anymore), so this feature have been removed. We plan to introduce a way to track executions for all functions after MVP (cc: @davidgamez)

infra/workflows/main.tf
```diff
@@ -107,5 +108,36 @@ resource "google_eventarc_trigger" "gtfs_validator_trigger" {
 
 }
 
+# Trigger 2: Trigger the workflow when a dataset is updated
```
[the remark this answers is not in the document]
cka-y: In the spirit of making all this process less confusing, this trigger has been removed. I'll make sure to document all the triggers in #413.

functions-python/update_validation_report/src/main.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    Execute the workflow for the latest datasets that need their validation report to be updated
+    :param latest_datasets: List of tuples containing the feed stable id and dataset stable id
+    :return: List of dataset stable ids for which the workflow was executed
+    """
+    project_id = f"mobility-feeds-{env}"
+    location = os.getenv("LOCATION", "northamerica-northeast1")
+    execution_triggered_datasets = []
+    batch_size = int(os.getenv("BATCH_SIZE", 5))
+    sleep_time = int(os.getenv("SLEEP_TIME", 5))
```
jcpitre, line 198: Why 5 seconds?

jcpitre (approved):
I'm always a bit weary of unconditional sleeps, but it works and that's what is important at this point.
Good job!

functions-python/update_validation_report/src/main.py
```diff
@@ -0,0 +1,227 @@
...
+            logging.info(f"Executing workflow for {feed_id}/{dataset_id}")
+            execute_workflow(project_id, input_data=input_data)
+            execution_triggered_datasets.append(dataset_id)
+        except Exception as e:
+            logging.error(
+                f"Error while executing workflow for {feed_id}/{dataset_id}: {e}"
+            )
+        count += 1
+        if count % batch_size == 0:
+            logging.info(
```
davidgamez, line 223: [suggestion]: add the count and len(latest_datasets) to the log in case anyone is following up on the process.

cka-y: ✅ done

Issue #353, opened 2024-04-01
Develop GTFS Realtime feed page


emmambd:
## Tasks

- [ ] [Match GTFS Realtime individual feed page](https://www.figma.com/file/LZVQecGplb4Lnuh7a24Jef/GTFS-Tool-Forms?type=design&node-id=1254%3A1743&mode=design&t=ztQe5LNqVxOuqd8y-1). Using v1/gtfs_rt_feeds/id
- [ ]  Entity type should be transformed to: 

          sa = Service Alerts
          tu = Trip Updates
          vp = Vehicle Positions

- [ ] "Auth Parameter" should have a different prefix based on the authentication type: 

            1 = API Key
            2 = HTTP

## Context
- Open the Data Sync plugin in Figma to see related fields in the API
<img width="303" alt="Screenshot 2024-04-05 at 10 47 14 AM" src="https://github.com/MobilityData/mobility-feed-api/assets/90632321/1971b961-9c5a-4f18-b568-93e9787e9f8f">

- All null values should not be displayed.

PR #425, opened 2024-05-08
feat: add single GTFS Feed and GTFS-RT Feed pages 

### LINKED ISSUE #352: Develop GTFS Schedule feed page
- [ ] Match individual GTFS Schedule feed page to [design ](https://www.figma.com/file/LZVQecGplb4Lnuh7a24Jef/GTFS-Tool-Forms?type=design&node-id=1063-857&mode=dev). Using /v1/gtfs_feeds/id for most of the page and v1/gtfs_feeds/id/datasets for Previous Datasets section.

- [ ] Bounding box visual computed with bounding box lat longs from v1/gtfs_feeds/id.  OpenStreetMap is used in this example. Should have zoom option. Open to other libraries depending on feedback.

- [ ] Data quality information extracted from validation_report object. 

- [ ] "Auth Parameter" should have a different prefix based on the authentication type: 

            1 = API Key
            2 = HTTP

- [ ] Feature list should display based on which features appear in the `validation_report` object for the feed. [Full list is here.](https://docs.google.com/document/d/1GGHEo8_RcvPBx74Y9HD2rXC5OdLpQplElMBDADwFC40/edit)

## Context
- Open the Data Sync plugin in Figma to see related fields in the API

Screenshot 2024-04-05 at 12 20 55 PM

- All null values should not be displayed.

indraneel:
**Summary:**

Creates `Feed.tsx` to match individual GTFS Schedule feed page to [design ](https://www.figma.com/file/LZVQecGplb4Lnuh7a24Jef/GTFS-Tool-Forms?type=design&node-id=1063-857&mode=dev). and match [Match GTFS Realtime individual feed page](https://www.figma.com/file/LZVQecGplb4Lnuh7a24Jef/GTFS-Tool-Forms?type=design&node-id=1254%3A1743&mode=design&t=ztQe5LNqVxOuqd8y-1). Closes #352 and #353

**Expected behavior:** 

image

image

**Testing tips:**

Navigate manually to `/feeds/<mdb-##>`


web-app/src/app/screens/Feed.tsx
```diff
@@ -0,0 +1,603 @@
...
+                            <TableRow>
+                              <TableCell>
+                                <b>Feature</b>
+                              </TableCell>
+                            </TableRow>
+                          )}
+                        {latestDataset?.validation_report?.features?.map(
+                          (v) => (
+                            <TableRow key={v}>
+                              <TableCell>{v}</TableCell>
```
Alessandro100, line 465: format feature name using `DATASET_FEATURES_FILES_MAPPING`  + add second column of "File or Field Associated"

web-app/src/app/screens/Feed.tsx
Alessandro100, line 1: This file is quite large, it would be great to see it broken down into smaller components

Alessandro100, line 1: We should handle the error state if a call fails, do you want to add it in this PR or create a new ticket / task?

web-app/src/app/screens/Feed.tsx
indraneel: Based on my conversation with @emmambd, I'm under the impression we are omitting "File or Field Associated"

emmambd: Correct, but the feature names themselves will be used

web-app/src/app/screens/Feed.tsx
indraneel: Sure - do you have a sense of how you might prefer this to be split? Logically I can imagine splitting the render function into:
- `renderGtfsFeedPage`
- `renderGtfsRTFeedPage`

indraneel: I can handle a basic error state here. If we need more complex design / changes I think that would be better suited for a different PR.

Alessandro100: I see that GtfsFeed and RT share a fair amount of code, maybe we can have components for elements on the page `FeedSummary`, `Datasets` etc

Alessandro100: a basic error state is perfect for now. @emmambd do you see a world where we have specific error states ex: "Error making the http call" and "Error: This feed Id does not exist" or just general "Error, go back home"

emmambd: @Alessandro100 I think basic error state is sufficient for now. I can't think of a common use case where we'd need something more complex in a situation where the feed itself won't display.

indraneel: added a basic error state w/ the following copy
image

indraneel: Pushed changes with functions that render the following:

- `renderFeedSummary`
- `renderDataQualitySummary`
- `renderFeaturesList`
- `renderAssociatedGTFSFeed`
- `renderPreviousDatasets`

web-app/src/app/screens/Feed.tsx
indraneel: For `mdb-123` does this look correct? 
image

web-app/src/app/screens/Feed.tsx
```diff
@@ -0,0 +1,561 @@
...
+                    </a>
+                  </Button>
+                )}
+                <Button
+                  variant='contained'
+                  sx={{ m: 2 }}
+                  endIcon={<LaunchOutlined />}
+                >
+                  <a
+                    href={feed?.source_info?.license_url}
```
Alessandro100, line 496: If the `license_url` is empty, I would just hide this button

web-app/src/app/screens/Feed.tsx
```diff
@@ -0,0 +1,561 @@
...
+                    See License
+                  </a>
+                </Button>
+                <Button
+                  variant='contained'
+                  sx={{ m: 2 }}
+                  endIcon={<LaunchOutlined />}
+                >
+                  <a
+                    href={feed?.source_info?.authentication_info_url}
```
Alessandro100, line 510: If `authentication_info_url` is empty I would hide this button as well

web-app/src/app/screens/Feed.tsx
```diff
@@ -0,0 +1,561 @@
...
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    if (user?.accessToken !== undefined && feedId !== undefined) {
+      dispatch(loadingFeed({ feedId, accessToken: user?.accessToken }));
+      dispatch(loadingDataset({ feedId, accessToken: user?.accessToken }));
+    }
+  }, []);
+
+  const renderFeedSummary = (): JSX.Element => (
```
Alessandro100, line 73: I love the methods made to render this content, it makes the code much more structured. I'd like to see these methods in their own component to bring this file from ~600 lines to ~250 lines

indraneel: Pushed this change.

indraneel:
> @indraneel It looks like there's some issues with the realtime vs. schedule querying - I can see the page load a realtime page before going to the GTFS Schedule feed:
> 
>  Screen.Recording.2024-05-22.at.8.44.38.AM.mov 
> It also looks like when a URL doesn't exist it's defaulting to some kind of empty link - can we remove the fields when the URL isn't present?
> 
>  Screen.Recording.2024-05-22.at.8.47.43.AM.mov

I pushed changes addressing these issues. Thanks  for spotting them @emmambd

Alessandro100 (approved):
Awesome changes, this looks great! Just merge the latest from main and it's good to go!

emmambd:
Looks like the feature flag wasn't set here, so the empty feeds page is on prod. #477. Should use and check for feature flags being enabled for front end work after this.

Issue #445, opened 2024-05-23
Default sort isn't by mdb-id + limit intermittently works


emmambd:
### Describe the bug

1. Setting a limit on /gtfs-feeds and /feeds works sometimes works and sometimes doesn't. It may be difficult to accurately replicate this but steps below:
2. Default sort for /gtfs-feeds and /feeds is not numeric by mdb-id

### Steps/Code to Reproduce

**Limit not working**

1. Run 

```
curl -X 'GET' \
  'https://api.mobilitydatabase.org/v1/gtfs_feeds?limit=10' \
  -H 'accept: application/json' \
[...]

PR #460, opened 2024-06-06
fix: default sort isn't by mdb-id + limit intermittently works

davidgamez:
## Summary:

This PR fixes the limit and offsets issues for endpoints `feeds`. The _"previous"_ implementation uses joins to bring entities from related tables. As there are a couple of x-to-many relations, the resulting data contained multiple rows from the same feed. Example datasets, locations, and redirects. One solution explorer was to apply limits and offsets in memory, but it was discarded due to the inefficiency of loading data upfront and making it hard to apply sorting. 
The solution proposed in the PR removes the need for joins, as Alchemy 2.x will do the join and load data when necessary. This makes the implementation simpler, see snipped below to follow a general pattern:
```python
# Create query filter
        feed_filter = FeedFilter(status=status, provider__ilike=provider, producer_url__ilike=producer_url)
# Base query with just the entity type
        feed_query = feed_filter.filter(Database().get_query_model(Feed))
# Apply limit and offset to the query. At this point, no data is retrieved yet, this is just a query change
        if limit is not None:
            feed_query = feed_query.limit(limit)
        if offset is not None:
            feed_query = feed_query.offset(offset)
# Retrieve all data applying limit, offset, and filters at database level
        results = feed_query.all()
# Use transformers to get the API response from the ORM instances
        return [BasicFeedImpl.from_orm(feed) for feed in results]
```
 As a side effect the refactor affected `feeds/:id`, this endpoint need to be tested too.
 As part of this PR the `v1/feeds` response is sorted by provider and stable_id.

## Observations
 Cases like the [latest_dataset transformation](https://github.com/MobilityData/mobility-feed-api/blob/8b650e6ac0c79c73d65812a94cb601f33f7b3c50/api/src/feeds/impl/models/gtfs_feed_impl.py#L27) might be a performance risk, as the code iterates over all datasets to filter the latest one.  This is done as the transformers don't have access to the DB and only transform what is provided in the `from_orm` function's parameter. Let's monitor this and make future changes to optimize it.  We need to keep a balance between easy-to-implement solutions and performance optimization.

Part of: #445 

## Out of Scope
 - `/v1/gtfs_feeds`
 - `v1/gtfs_rt_feeds`

**Expected behavior:** 

The limit parameter works as expected. 

**Testing tips:**

 Local testing:
- Setup your local DB and add multiple datasets to multiple feeds.
- Execute:
```
curl --location 'http://localhost:8080/v1/feeds?limit=10' \
--header 'Accept: application/json'
```
- Expect the limit parameter to be respected
- Test with multiple combinations of limit and offset parameters

### DIFF
--- a/api/src/database/database.py
+++ b/api/src/database/database.py
@@ -5,7 +5,7 @@ import uuid
 from typing import Type, Callable
 from dotenv import load_dotenv
 from sqlalchemy import create_engine, inspect
-from sqlalchemy.orm import load_only, Query, class_mapper
+from sqlalchemy.orm import load_only, Query, class_mapper, Session
 
 from database_gen.sqlacodegen_models import Base, Feed, Gtfsfeed, Gtfsrealtimefeed
 from sqlalchemy.orm import sessionmaker
@@ -183,6 +183,19 @@ class Database:
         finally:
             if update_session:
                 self.close_session()
+
+    def get_session(self) -> Session:
+        """
+        :return: the current session
+        """
+        return self.session
+
+    def get_query_model(self, model: Type[Base]) -> Query:
+        """
+        :param model: the sqlalchemy model to query
+        :return: the query model
+        """
+        return self.get_session().query(model)
 
     def select_from_active_session(self, model: Base, conditions: list = None, attributes: list = None):
         """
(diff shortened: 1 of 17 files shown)


qcdyx (approved):
I've tested the limit with my local db, and it works perfectly.

PR #474, opened 2024-06-13
feat: feeds + search page

### LINKED ISSUE #348: Create Feeds page and update navigation
## Definition of done

- Create [Feeds page to match design](https://www.figma.com/file/LZVQecGplb4Lnuh7a24Jef/GTFS-Tool-Forms?type=design&node-id=1193%3A1431&mode=design&t=cu4QL4TfYjrF7ETJ-1)
- Sorted by default as GTFS Schedule data type first, in order of most recent downloaded_at
- Provider field should be restricted to 100 character display. 

## Context

- Will be able to view which API fields correspond to which objects in Figma using the Data Sync plugin
- Adding dynamic data is blocked by #350. To be done in #351.

indraneel:
**Summary:**

This is to close #348. 

Navigate to `/feeds` and type your search query in the search bar. 

image


web-app/src/app/screens/Feeds/index.tsx
```diff
@@ -0,0 +1,265 @@
...
+            },
+          }),
+        );
+      }
+    }, 400);
+
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [searchLimit, searchParams, selectedFeedTypes]);
```
davidgamez, line 112: Let's add here the login event status. This can remove the need of the timer. Example code: https://github.com/MobilityData/mobility-feed-api/blob/8c92065d942993fcdcb5531f8b313fd21da723d3/web-app/src/app/screens/Feed/index.tsx#L73

indraneel: I see you added this in main and I merged it in. The timer is needed to prevent search API calls on each keystroke, but this is probably better done by debouncing which I can add.

davidgamez: The _isAuthenticatedOrAnonymous_ is needed to avoid calls to the API that returned 403 as no users/anonymous cannot be in the initial page rendering state. I got the idea of the timer now, thanks. We were thinking of a different user interaction, but the design will be available for review later this week.

web-app/src/app/store/saga/feeds-saga.ts
```diff
@@ -0,0 +1,29 @@
...
+import { searchFeeds } from '../../services/feeds';
+
+function* getFeedsSaga({
+  payload: { params, accessToken },
+}: PayloadAction<{
+  params: AllFeedsParams;
+  accessToken: string;
+}>): Generator<StrictEffect, void, AllFeedsType> {
+  try {
+    // debugger;
```
Alessandro100, line 19: we can remove this comment

Alessandro100:
Code looks good and works well! Something I found is when we click "Load 20 more feeds" the current behaviour is that the page seems to do a refresh and I'm brought back to the top. The expected behaviour is that after your click the button the page stays in the same location with more elements appearing. If you think this is a quick fix we can do it in this ticket, otherwise we can create another ticket

Alessandro100:
Another usability issue found
Steps
1. Go on feeds page and do a search
2. Go to another page and then back to the feeds page
3. The search bar will be empty, the url query params will be empty, but the previous search results will remain

expected behavior: If there is no search values, there should be no displayed search results

Alessandro100:
> Code looks good and works well! Something I found is when we click "Load 20 more feeds" the current behaviour is that the page seems to do a refresh and I'm brought back to the top. The expected behaviour is that after your click the button the page stays in the same location with more elements appearing. If you think this is a quick fix we can do it in this ticket, otherwise we can create another ticket

Ignore this comment, functionality will change and we are considering pagination

indraneel:
> Reset feeds when navigating away

[Addressed in this change](https://github.com/MobilityData/mobility-feed-api/pull/474/commits/b9fa41670f14a55553c087bc1da8eef1ebfb39da)

Alessandro100 (approved):
looks good, just bring it up to date with main and it will be good to go

Alessandro100:
@stefan-philip would you be able to sign the CLA :)

PR #501, opened 2024-06-19
feat: validate downloaded files as ZIP before processing dataset

cka-y:
**Summary:**

This update enhances batch processing by validating that the producer URL returns a zipped file before saving to GCP and updating the database. As part of this improvement, I also removed from GCP and the database all datasets and related entities that were not zipped files and not linked to other records. Below are the details of all invalid datasets per environment:

- [invalid_datasets_prod.csv](https://github.com/user-attachments/files/15904126/invalid_datasets_prod.csv) $\approx$ 25% of all 8609 datasets are invalid zips.
- [invalid_datasets_dev.csv](https://github.com/user-attachments/files/15904125/invalid_datasets_dev.csv)
- [invalid_datasets_qa.csv](https://github.com/user-attachments/files/15904127/invalid_datasets_qa.csv)

**Expected Behavior:**

The system should check if the producer URL returns a zipped file. If the URL does not return a zipped file, the dataset should not be stored in GCP or persisted in the database.

**Testing Tips:**
- Adjust the `FEED_LIMIT` variable in `batch-datasets` to cover a wide range of datasets (default is 10).
- Run the function using the following command:
```bash
curl -m 30 -X POST https://northamerica-northeast1-mobility-feeds-dev.cloudfunctions.net/batch-datasets-dev \
-H "Content-Type: application/json"
```
- Inspect the logs to ensure that each URL identified as not being a zip file is indeed not a zip file. You can filter the logs using the search string `is not a valid ZIP file`, which is logged as an `ERROR` level message in the `batch-process-dataset` cloud function.
- Verify that invalid datasets are correctly logged, reported, and omitted from GCP and the database.
 
### DIFF
--- a/functions-python/helpers/utils.py
+++ b/functions-python/helpers/utils.py
@@ -15,6 +15,7 @@
 #
 
 import os
+import zipfile
 
 from google.cloud import storage
 from requests.adapters import HTTPAdapter
@@ -126,3 +127,15 @@ def download_and_get_hash(
         if os.path.exists(file_path):
             os.remove(file_path)
         raise e
+
+
+def is_zip_file(file_path):
+    """
+    Checks if the file is a valid ZIP file.
+    :param file_path: the path to the file
+    """
+    try:
+        with zipfile.ZipFile(file_path, "r"):
+            return True
+    except zipfile.BadZipFile:
+        return False
--- a/functions-python/helpers/utils.py
+++ b/functions-python/helpers/utils.py
@@ -14,15 +14,14 @@
 #  limitations under the License.
 #
 
+import hashlib
 import os
-import zipfile
 
+import requests
+import urllib3
 from google.cloud import storage
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
-import requests
-import hashlib
-import urllib3
 from urllib3.util.ssl_ import create_urllib3_context
 
 
@@ -127,15 +126,3 @@ def download_and_get_hash(
         if os.path.exists(file_path):
             os.remove(file_path)
         raise e
-
-
-def is_zip_file(file_path):
-    """
-    Checks if the file is a valid ZIP file.
-    :param file_path: the path to the file
-    """
-    try:
-        with zipfile.ZipFile(file_path, "r"):
-            return True
-    except zipfile.BadZipFile:
-        return False
--- a/functions-python/batch_process_dataset/src/main.py
+++ b/functions-python/batch_process_dataset/src/main.py
@@ -19,6 +19,7 @@ import json
 import os
 import random
 import uuid
+import zipfile
 from dataclasses import dataclass
 from datetime import datetime
 from typing import Optional
@@ -39,7 +40,7 @@ from helpers.database import (
 import logging
 
 from helpers.logger import Logger
-from helpers.utils import download_and_get_hash, is_zip_file
+from helpers.utils import download_and_get_hash
 
 
 @dataclass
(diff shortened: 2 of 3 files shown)


cka-y:
Most skipped datasets are not valid ZIP files. However, there are a few cases where accessing the `producer_url` via a web browser successfully downloads a valid ZIP file, but programmatically attempting to replicate this process results in an HTML file instead. An example for `mdb-683` returns:

```html
<html>
<head><title>410 Gone</title></head>
<body>
<center><h1>410 Gone</h1></center>
<hr><center>openresty</center>
</body>
</html>
```

Here is a list of stable IDs and producer URLs where this issue is known to occur (note that there may be others):

- [mdb-232](https://www.fnsb.gov/DocumentCenter/View/690/General-Transit-Feed-Specification-ZIP) (returns an empty response when accessed programatically)
- [mdb-919](https://www.cityoflompoc.com/home/showdocument?id=29896) (returns an HTML file when accessed programatically)
- [mdb-683](https://dipl.nt.gov.au/data-feeds/bus-gtfs/google-transit-alice-springs.zip?v=0.7.0) (returns an HTML file when accessed programatically)

cka-y:
Here is the list of feeds in PROD with no datasets. Note that this list might reduce if we run the `batch-datasets` function. However, we should not run it before updating the `batch-process-dataset` to ignore invalid ZIP files; otherwise, I'll need to restart the process. (cc: @emmambd)

[prod-feeds-w-no-dataset.csv](https://github.com/user-attachments/files/15904929/prod-feeds-w-no-dataset.csv)

cka-y:
To avoid manually deleting files and database entities again, we should consider the following options:

- **Option A:** Pause the `batch-datasets` scheduler in PROD to prevent storing new datasets, some of which might be invalid ZIP files.
- **Option B:** Update the `batch-process-datasets` function in PROD by running the `datasets-batch-deployer-prod` GitHub Action without releasing (since we are not ready for a full release).

I recommend Option A, depending on how long we need before releasing. However, if the scheduler is paused for too long, we might miss a significant number of historical datasets. Thoughts? cc: @emmambd @davidgamez

davidgamez:
> To avoid manually deleting files and database entities again, we should consider the following options:
> 
> * **Option A:** Pause the `batch-datasets` scheduler in PROD to prevent storing new datasets, some of which might be invalid ZIP files.
> * **Option B:** Update the `batch-process-datasets` function in PROD by running the `datasets-batch-deployer-prod` GitHub Action without releasing (since we are not ready for a full release).
> 
> I recommend Option A, depending on how long we need before releasing. However, if the scheduler is paused for too long, we might miss a significant number of historical datasets. Thoughts? cc: @emmambd @davidgamez

I prefer Option A. We can connect tomorrow at stand-up to make a decision.

functions-python/helpers/utils.py
```diff
@@ -126,3 +127,15 @@ def download_and_get_hash(
...
         raise e
+
+
+def is_zip_file(file_path):
+    """
+    Checks if the file is a valid ZIP file.
+    :param file_path: the path to the file
+    """
+    try:
+        with zipfile.ZipFile(file_path, "r"):
```
davidgamez, line 138: [question]: why not use https://docs.python.org/3/library/zipfile.html#zipfile.is_zipfile function instead?

emmambd:
Option A worries me significantly if we're waiting a full month. Let's discuss.

emmambd:
@cka-y based on your CSV, it looks like there's about 200 feeds that don't have any real value (don't return a dataset). From a quick glance, I think half of these probably are legitimately "deprecated" (meaning I can find a replacement feed for them), but there will also be some where we have no idea if the feed has truly been replaced or if it's just down. 

We might need an additional issue to return some kind of "non-fetchable" message in the API response and in the website, for this use case.

functions-python/helpers/utils.py
cka-y: didn't know it existed 🤭 -- i have updated the code

jcpitre (approved):

davidgamez (approved):
LGTM

emmambd:
Created #506 and #505 to address what's out of scope for this PR

Issue #420, opened 2024-05-07
Add associated realtime feeds to GTFS Schedule response


emmambd:
We include the static reference for a realtime feed in the Mobility Database, but currently we have no way to display which realtime feeds are associated with a GTFS Schedule feed. 

# Requirements
- Add the related realtime feeds to the gtfs endpoint.

UI story related: #467

Issue #467, opened 2024-06-12
Final GTFS Realtime feed page: Related Realtime Feeds


emmambd:
Describe the problem
[Final GTFS Realtime feed page design is here](https://www.figma.com/design/LZVQecGplb4Lnuh7a24Jef/GTFS-Tool-Forms?node-id=1370-1781&m=dev)

Proposed solution: 

Please style the following sections to match the composite:

- [ ] Related Realtime Feeds 
- [ ] Call `gtfs_feeds/{id}/gtfs_rt_feeds` to retrieve a list of associated realtime feeds

These realtime feeds should be pulled based on their static reference. Any realtime feed with the same static reference as the feed displayed on the webpage should be included within this component. 

Cypress tests should be included for these sections within the PR.

indraneel:
Hey @emmambd what does `static reference` refer to here?

indraneel:
This is partially blocked on #420 being complete (currently the API doesn't return associated realtime feeds)

emmambd:
@indraneel This would be a good small issue to work on before jumping back in the feed form.

PR #509, opened 2024-06-26
feat: add associated realtime feeds to GTFS Schedule response(OpenApi)

davidgamez:
**Summary:**

Part of #420 
Added `gtfs_rt_feeds` to the GTFS OpenApi entity. This affects endpoints: `gtfs_feeds`, `gtfs_feeds/{id}` and `search`

**Expected behavior:** 
The gtfs_rt_feeds field is added to the OpenApi model. Implementation is out of the scope of this PR.
Example of the _"related gtfs rt feed"_
```
...
"gtfs_rt_feeds": [
      {
        "id": "mdb-1210",
        "status": "deprecated",
        "provider": "Los Angeles Department of Transportation (LADOT, DASH, Commuter Express)",
        "feed_name": "Bus",
        "source_info": {
          "producer_url": "https://ladotbus.com/gtfs",
          "authentication_type": 2,
          "authentication_info_url": "https://apidevelopers.ladottransit.com",
          "api_key_parameter_name": "Ocp-Apim-Subscription-Key",
          "license_url": "https://www.ladottransit.com/dla.html"
        },
        "entity_types": [
          "vp"
        ]
      }
...
```

**Testing tips:**


docs/DatabaseCatalogAPI.yaml
```diff
@@ -367,10 +367,62 @@ components:
           properties:
             locations:
               $ref: "#/components/schemas/Locations"
-        - type: object
-          properties:
             latest_dataset:
               $ref: "#/components/schemas/LatestDataset"
+            gtfs_rt_feeds:
```
jcpitre, line 372: It seems to me that the feed response is becoming quite _fat_
Have you considered using the search or gtfs_rt_feeds endpoint to obtain this info?
It would require multiple calls to the API to obtain all the data, but don't we have to draw the line somewhere?

docs/DatabaseCatalogAPI.yaml
```diff
@@ -367,10 +367,62 @@ components:
...
+              type: array
+              items:
+                $ref: "#/components/schemas/BasicGtfsRTFeed"
+
+    BasicGtfsRTFeed:
```
jcpitre, line 377: Why not use BasicFeed as a "parent"?

davidgamez: I was thinking of adding BasicFeed, but BasicFeed has a lot of information that might not be necessary and can be retrieved if the consumer needs it. I'm happy to review this. What do you think?

docs/DatabaseCatalogAPI.yaml
davidgamez: The `search` endpoint uses the GtfsFeed entity in the response, but there is no plan to add it as a parameter.  Adding a parameter to the search might be less intuitive to discover than a field within the gtfs entity. 
We can filter by feed_id in the `gtfs_rt_feeds`, but this will be a different issue than the one we are implementing, see #420.  Regarding multiple calls, the UI at #467 will have to do "multiple" anyway, as the GTFS_RT screen will have to call the `gtfs_feeds/{id}` to retrieve the "related" realtime feeds.

docs/DatabaseCatalogAPI.yaml
davidgamez: I forgot to mention that BasicFeed doesn't contain `entity_types`.

docs/DatabaseCatalogAPI.yaml
jcpitre: But my main point is that the gtfsfeed response is getting quite big. 
You don't think that adding an argument to /v1/gtfs_rt_feeds like schedule_feed=mdb-1 would be better?
It would require calling this endpoint for every schedule feed we want to list, which is not great I guess.

docs/DatabaseCatalogAPI.yaml
jcpitre: Fine. GtfsRTFeed does contain the entity_type though.
In any case we can live with the duplications for now.

docs/DatabaseCatalogAPI.yaml
davidgamez: The requirement is to get the real-time feeds given a gtfs ID. Adding an argument will do, but it will be prompt to errors from the consumer side as the filtering will apply all the possible arguments. For more strict implementation, if the main concern is the size of the response, I would prefer to add an endpoint to express only the related feeds with path: `v1/gtfs_feeds/{id}/gtfs_rt_feeds`. This can return the complete real-time feed structure.

jcpitre: The way you propose is fine. To me it's more important to decide if we should return this info with the GtfsFeed response or require an extra call to the API to obtain it. I am leaning towards having the extra call.

davidgamez: > The way you propose is fine. To me it's more important to decide if we should return this info with the GtfsFeed response or require an extra call to the API to obtain it. I am leaning towards having the extra call. But I will leave the final decision to you.

In case of a new endpoint `v1/gtfs_feeds/{id}/gtfs_rt_feeds`, this will be a new call as it will return a full GTFS_rt instead of GTFS. So there is no additional entity to add to the model.

docs/DatabaseCatalogAPI.yaml
```diff
@@ -367,8 +385,6 @@ components:
           properties:
             locations:
               $ref: "#/components/schemas/Locations"
-        - type: object
-          properties:
             latest_dataset:
```
davidgamez, line 388: Unrelated to this PR. I fixed this "extra" type declaration as a consumer of the file might have issues rendering the yaml file or generating code.

docs/DatabaseCatalogAPI.yaml
davidgamez: @jcpitre Added a new endpoint and removed all the previous entities. It looks cleaner now!

docs/DatabaseCatalogAPI.yaml
```diff
@@ -196,6 +196,24 @@ paths:
               schema:
                 $ref: "#/components/schemas/GtfsDatasets"
 
+  /v1/gtfs_feeds/{id}/gtfs_rt_feeds:
+    parameters:
+      - $ref: "#/components/parameters/feed_id_of_datasets_path_param"
```
jcpitre, line 201: I'm being picky, but the description of `feed_id_of_datasets_path_param` says: `The ID of the feed for which to obtain datasets` which is not quite right.
Maybe you could use [feed_id_path_param](https://github.com/MobilityData/mobility-feed-api/blame/5b3c60410bae78475d95982c7e023d5789eb456e/docs/DatabaseCatalogAPI.yaml#L928C5-L928C23) instead?

jcpitre (approved):
LGTM!

PR #515, opened 2024-06-28
poc: recognizing when a dataset was unfetchable

cka-y:
**Summary:**
This PR is a proof of concept (POC) to identify when a dataset was not fetched, providing clearer information to the user about when and why a dataset was never fetched.

**Expected Behavior:**
Currently, the batch processing job attempts to fetch all non-deprecated feeds. Known reasons for a non-deprecated feed not being successfully fetched include:

1. Internal server errors (e.g., HTTP connection timeout because the URL is inaccessible).
2. Invalid ZIP file: If the dataset available at the producer's URL is not a valid ZIP file (e.g., the URL redirects to an HTML page), processing is skipped.

The goal of this POC is to enhance the UI by providing more information when a feed has no datasets. To achieve this, the existing `dataset_trace` GCP Datastore entity has been updated with an `INVALID_ZIP` status type. The status types are now as follows:
- `FETCHED`: The dataset was successfully updated, stored in GCP, and added to the database.
- `NOT_FETCHED`: There was no update; the latest stored dataset is still up to date.
- `FAILURE`: An error occurred during processing (e.g., HTTP connection timeout due to an inaccessible producer URL).
- `INVALID_ZIP_FILE`: The URL is accessible, but the response is not a valid ZIP file.

By having this status in Datastore, along with a `timestamp` and a `stable_id` corresponding to the feed's `stable_id`, it is possible to retrieve the latest status of retrieval for a given feed. Consequently, an element has been added to the API's GTFSFeed response to include information about the last fetch attempt. Here is an example of the object:
```json
"last_fetch_attempt": {
        "status": "PUBLISHED",
        "timestamp": "2024-06-19T17:40:50.188550"
}
```

This information is now displayed on the UI's feed pages to show the latest status. Here are a few examples for different types of statuses (note that the results may vary if the API development is rebuilt or the batch processing jobs are run using the preview link):
image
Screenshot 2024-06-28 at 2 10 49 PM
image
image

**Not covered by this POC**:
- In the final screenshot, you can see that the status is displayed as "in about 4 hours." This discrepancy is due to a timezone difference between the UI and the backend. The processing was completed recently on the backend, but the UI is operating in a different timezone, causing the time mismatch.

### DIFF
--- a/api/.openapi-generator/FILES
+++ b/api/.openapi-generator/FILES
@@ -14,9 +14,11 @@ src/feeds_gen/models/basic_feed.py
 src/feeds_gen/models/bounding_box.py
 src/feeds_gen/models/external_id.py
 src/feeds_gen/models/extra_models.py
+src/feeds_gen/models/fetch_status.py
 src/feeds_gen/models/gtfs_dataset.py
 src/feeds_gen/models/gtfs_feed.py
 src/feeds_gen/models/gtfs_rt_feed.py
+src/feeds_gen/models/last_fetch_attempt.py
 src/feeds_gen/models/latest_dataset.py
 src/feeds_gen/models/latest_dataset_validation_report.py
 src/feeds_gen/models/location.py
--- a/api/.openapi-generator/FILES
+++ b/api/.openapi-generator/FILES
@@ -14,7 +14,6 @@ src/feeds_gen/models/basic_feed.py
 src/feeds_gen/models/bounding_box.py
 src/feeds_gen/models/external_id.py
 src/feeds_gen/models/extra_models.py
-src/feeds_gen/models/fetch_status.py
 src/feeds_gen/models/gtfs_dataset.py
 src/feeds_gen/models/gtfs_feed.py
 src/feeds_gen/models/gtfs_rt_feed.py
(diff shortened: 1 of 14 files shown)


davidgamez:
Nice work! Leaving here a few notes after connecting with @cka-y 
- We need to be aware of Datastore scanning all records as part of the queries(_this is why the current POC tries to get a timestamp to filter out records as part of the query_)
- Having this information as part of the _"common"_ GTFS entity will affect all endpoints to the detriment of performance.
- If we implement a _"generic"_ logging system, it will allow us to add additional "operations" like validator-runs, bounding-box extraction, etc.
- We agreed that it's preferable to add this information as a separate endpoint rather than in the GTFS entity. Example proposed endpoint: `v1/gtfs_feeds/{id}/operations_trace?operation=FETCH_DATASET`.

Next steps:
- [ ] Align UI layout design and create a UI issue
- [ ] Create API issue to add the new endpoint

emmambd:
This is amazing! Thoughts on making this work a stretch goal as part of our August wrap-up of the Mobility Database UI? This is super high value, but I want to be mindful that we should start the Flex rules work next sprint and this isn't critical for beta or launch.

cka-y:
@emmambd, I believe we can include this task on the board as a low priority item. Implementing it shouldn't be too difficult if we break it down into separate backend and frontend tasks. If the current design doesn't meet your expectations, we would need your input on how and where you would like the information displayed in the UI.

cka-y:
Closing this 
Follow up issues: #529 and #530

Issue #516, opened 2024-06-28
Lack a global mechanism for managing user states during cold loading


qcdyx:
### Describe the bug

Continues #443, the web app lacks a global mechanism for managing user states. When a user opens a static page (About, Privacy Policy, or Terms and Conditions) in a new tab, the app does not manage the user state.

If a user is already logged in, and clicked on the bookmarked login address, the app returns a blank page.
![image](https://github.com/MobilityData/mobility-feed-api/assets/5789435/dd2baad1-a9b9-4b62-a055-906623ca245c)

I suggest manager user state/refresh user in the header.tsx.

### Steps/Code to Reproduce

1. Login
2. Open Privacy Policy/login in a new tab, user state is undefined.

### Expected Results

User should stay logged in (User state is managed globally).

### Actual Results

User state is undefined.

qcdyx:
As discussed in the standup, we need to automatically generate an access token upon user login and refresh it every hour.

PR #528, opened 2024-07-04
fix: use firebase access token directly instead of redux saved token

davidgamez:
**Summary:**

This PR addresses using the user's access token while requesting the feed's API endpoints. Principles considered for using the Firebase SDK function instead of the redux state for the API token management:

- Previously the access token used is the one saved in the redux store. This token was initially saved in redux as part of the account screen. Using this token can lead to cases where the token is already created before the user hits the account screen, having the code to manage different entry stages to the create access token button. In addition, access tokens expire every hour, so we will need to handle the rotation of the token and the implications of this process on the current account's page. There is a potential race condition when updating a token to the redux state, and the user requests an API call with an already expired token.
- The PR uses the token _managed_ by the Firebase SDK directly. The firebase already caches the token and is able to request a new one if it expires. The additional benefit is that the access token management is also abstracted from the UI component, leaving the service layer to perform token rotation when needed.

Part of #516 

## Out of scope

- This PR doesn't address the persistent of the redux state across tabs
- Cypress test will be added in a separate PR

**Expected behavior:** 

The search and feed screen shows the correct information for anonymous and authenticated users.

**Testing tips:**

Anonymous user:
- Open the App in the home screen
- Click "Browse Feeds"
- Expected feed results to render on the screen
- Click on a feed and verify that all information is presented
- Wait an +hour(do not refresh the page)
- Click back and verify that feed results are rendered

Authenticated user:
- Open the App in the home screen
- Login to the app with any method(please try multiple one at a time)
- Click the "Feeds" link from the top header menu
- Expected feed results to render on the screen

- Try also different combinations of login and logout operations

### DIFF
--- a/web-app/src/app/services/profile-service.ts
+++ b/web-app/src/app/services/profile-service.ts
@@ -35,6 +35,22 @@ export const getUserFromSession = async (): Promise<User | null> => {
   };
 };
 
+/**
+ * Retrieve the current Firebase user's Access token.
+ * If the access token is expired, this function will refresh it and return it.
+ * If the access token is not available, an error is thrown.
+ */
+export async function getUserAccessToken(): Promise<string> {
+  const currentUser = app.auth().currentUser;
+
+  if (currentUser === null) {
+    throw new Error('Cannot retrieve access token');
+  }
+  const idTokenResult = await currentUser.getIdTokenResult();
+
+  return idTokenResult.token;
+}
+
 export const generateUserAccessToken = async (
   user: User,
 ): Promise<User | null> => {
--- a/web-app/src/app/store/saga/dataset-saga.ts
+++ b/web-app/src/app/store/saga/dataset-saga.ts
@@ -13,9 +13,7 @@ function* getDatasetSaga({
 }: PayloadAction<{ feedId: string }>): Generator {
   try {
     if (feedId !== undefined) {
-      const accessToken: string | null = (yield call(
-        getUserAccessToken,
-      )) as string;
+      const accessToken = (yield call(getUserAccessToken)) as string;
       const datasets = (yield call(
         getGtfsFeedDatasets,
         feedId,
--- a/web-app/src/app/store/saga/dataset-saga.ts
+++ b/web-app/src/app/store/saga/dataset-saga.ts
@@ -1,4 +1,4 @@
-import { type StrictEffect, call, takeLatest, put } from 'redux-saga/effects';
+import { call, takeLatest, put } from 'redux-saga/effects';
 import { loadingFeedFail } from '../feed-reducer';
 import { getAppError } from '../../utils/error';
 import { DATASET_LOADING_FEED, type FeedError } from '../../types';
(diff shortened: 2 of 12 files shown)


web-app/src/app/store/saga/dataset-saga.ts
```diff
@@ -1,22 +1,27 @@
...
-  paths['/v1/gtfs_feeds/{id}/datasets']['get']['responses'][200]['content']['application/json']
-> {
+  payload: { feedId },
+}: PayloadAction<{ feedId: string }>): Generator {
   try {
     if (feedId !== undefined) {
-      const datasets = yield call(getGtfsFeedDatasets, feedId, accessToken, {});
+      const accessToken: string | null = (yield call(
+        getUserAccessToken,
+      )) as string;
```
Alessandro100, line 18: we don't need to declare `string | null` and `as string`

web-app/src/app/store/saga/feed-saga.ts
```diff
@@ -20,16 +20,20 @@ import {
...
+}: PayloadAction<{ feedId: string }>): Generator<
   StrictEffect,
   void,
   AllFeedType
 > {
   try {
     if (feedId !== undefined) {
+      const accessToken: string | null = (yield call(
+        getUserAccessToken,
+      )) as string;
```
Alessandro100, line 36: same comment

web-app/src/app/store/saga/feed-saga.ts
```diff
@@ -43,15 +47,14 @@ function* getFeedSaga({
...
-  AllFeedType[]
-> {
+  payload: { feedIds },
+}: PayloadAction<{ feedIds: string[] }>): Generator {
   try {
     if (feedIds.length > 0) {
-      const feedsData: AllFeedType[] = yield all(
+      const accessToken: string | null = (yield call(
+        getUserAccessToken,
+      )) as string;
```
Alessandro100, line 56: same

web-app/src/app/services/profile-service.ts
```diff
@@ -35,6 +35,22 @@ export const getUserFromSession = async (): Promise<User | null> => {
...
+export async function getUserAccessToken(): Promise<string> {
+  const currentUser = app.auth().currentUser;
+
+  if (currentUser === null) {
+    throw new Error('Cannot retrieve access token');
+  }
+  const idTokenResult = await currentUser.getIdTokenResult();
+
+  return idTokenResult.token;
+}
```
Alessandro100, line 52: this is great! Handles a decent amount of auth info for us 🎉

Alessandro100 (approved):
left a few small comments otherwise it works great ✅

web-app/src/app/store/saga/dataset-saga.ts
davidgamez: Good observation. However, it's actually needed at list the `as string`. The reason is that generators functions don't play nice with TS with multiple calls. The root cause is that the generators functions "always" returns any. we were enforcing the type discovery defined  StriictEffect type but this doesn't account for multiple calls. The way I was able to go around is to define the return type a generic `Generator` vs the previous pattern was like `Generator<StrictEffect, void,AllFeedType>`.  
In any case, you are right I can skip the `string | null` definition keeping the `as string`, resulting code `const accessToken = (yield call(getUserAccessToken)) as string;`

qcdyx (approved):
LGTM

davidgamez:
> left a few small comments otherwise it works great ✅

Thanks!! addressed your feedback!

Issue #537, opened 2024-07-08
304 error when returning feeds on QA


emmambd:
### Describe the bug

304 error when returning feeds on QA

### Steps/Code to Reproduce

304 error when returning feeds on QA

### Expected Results

304 error when returning feeds on QA

### Actual Results

304 error when returning feeds on QA

PR #538, opened 2024-07-08
load components when app is ready

davidgamez:
**Summary:**

Fixes: #537
The feed screen is not rendering after hitting refresh on the page.

## How to reproduce the issue:
- Open the app
- Click Browse
- Click the any feed and wait for the feed screen to be fully render
- Refresh page
- Current behavior => error raised with text : _There was an error loading the feed._
- Expected behavior => Feed screen is rendered correctly

**Expected behavior:** 

The feed screen is rendered after the page is refreshed.

**Testing tips:**

Follow the _How to reproduce the issue_. Test with different authentication state combinations.

### DIFF
--- a/web-app/src/app/store/profile-reducer.ts
+++ b/web-app/src/app/store/profile-reducer.ts
@@ -93,6 +93,7 @@ export const userProfileSlice = createSlice({
       state.status = 'unauthenticated';
       state.isSignedInWithProvider = false;
       state.isAppRefreshing = false;
+      state.user = undefined;
     },
     logoutFail: (state) => {
       state.status = 'unauthenticated';
--- a/web-app/src/app/store/saga/auth-saga.ts
+++ b/web-app/src/app/store/saga/auth-saga.ts
@@ -28,7 +28,6 @@ import {
   resetPasswordSuccess,
   verifyFail,
   verifySuccess,
-  anonymousLogin,
   anonymousLoginFailed,
 } from '../profile-reducer';
 import { type NavigateFunction } from 'react-router-dom';
@@ -80,10 +79,9 @@ function* logoutSaga({
   navigateTo: NavigateFunction;
 }>): Generator {
   try {
+    navigateTo(redirectScreen);
     yield app.auth().signOut();
     yield put(logoutSuccess());
-    yield put(anonymousLogin()); // Use anonymous login to keep the user signed in
-    navigateTo(redirectScreen);
   } catch (error) {
     yield put(loginFail(getAppError(error) as ProfileError));
   }
--- a/web-app/src/app/App.tsx
+++ b/web-app/src/app/App.tsx
@@ -8,13 +8,24 @@ import { RemoteConfigProvider } from './context/RemoteConfigProvider';
 import { useDispatch } from 'react-redux';
 import { anonymousLogin } from './store/profile-reducer';
 import i18n from '../i18n';
-import { Suspense, useEffect } from 'react';
+import { Suspense, useEffect, useState } from 'react';
 import { I18nextProvider } from 'react-i18next';
+import { app } from '../firebase';
 
 function App(): React.ReactElement {
   require('typeface-muli'); // Load font
   const dispatch = useDispatch();
+  const [isAppReady, setIsAppReady] = useState(false);
+
   useEffect(() => {
+    app.auth().onAuthStateChanged((user) => {
+      if (user != null) {
+        setIsAppReady(true);
+      } else {
+        setIsAppReady(false);
+        dispatch(anonymousLogin());
+      }
+    });
     dispatch(anonymousLogin());
   }, [dispatch]);
 
@@ -26,7 +37,7 @@ function App(): React.ReactElement {
             <AppSpinner>
               <BrowserRouter>
                 <Header />
-                <AppRouter />
+                {isAppReady ? <AppRouter /> : null}
               </BrowserRouter>
             </AppSpinner>
             <Footer />


web-app/src/app/store/profile-reducer.ts
```diff
@@ -93,6 +93,7 @@ export const userProfileSlice = createSlice({
       state.status = 'unauthenticated';
       state.isSignedInWithProvider = false;
       state.isAppRefreshing = false;
+      state.user = undefined;
```
davidgamez, line 96: This is not related to the fix. However, I noticed the user information stayed in the store after logout. This might lead to unexpected behavior after completing the logout and before the anonymous login succeeds.

web-app/src/app/store/saga/auth-saga.ts
```diff
@@ -80,10 +79,9 @@ function* logoutSaga({
   navigateTo: NavigateFunction;
 }>): Generator {
   try {
+    navigateTo(redirectScreen);
     yield app.auth().signOut();
     yield put(logoutSuccess());
-    yield put(anonymousLogin()); // Use anonymous login to keep the user signed in
```
davidgamez, line 85: Anonymous login is not needed anymore here. Now anonymous login is being centralized in the root component.

web-app/src/app/store/saga/auth-saga.ts
```diff
@@ -80,10 +79,9 @@ function* logoutSaga({
   navigateTo: NavigateFunction;
 }>): Generator {
   try {
+    navigateTo(redirectScreen);
```
davidgamez, line 82: We are navigating out of the current screen before logging out to avoid race conditions.

davidgamez:
Login and logout flow with multiple tabs open.
Uploading Screen Recording 2024-07-09 at 2.44.27 PM.mov…

web-app/src/app/components/LogoutConfirmModal.tsx
```diff
@@ -25,7 +25,9 @@ export default function ConfirmModal({
   const dispatch = useAppDispatch();
   const navigateTo = useNavigate();
   const confirmLogout = (): void => {
-    dispatch(logout({ redirectScreen: SIGN_IN_TARGET, navigateTo }));
+    dispatch(
+      logout({ redirectScreen: SIGN_OUT_TARGET, navigateTo, propagate: true }),
```
Alessandro100, line 29: [question] what does the `propagate` variable control?

web-app/src/app/router/Router.tsx
```diff
@@ -17,8 +17,34 @@ import TermsAndConditions from '../screens/TermsAndConditions';
...
+
+  const loginUser = (): void => {
+    // Refresh the page to ensure the user is authenticated
+    window.location.reload();
+  };
+
+  useEffect(() => {
+    createDispatchChannel(LOGOUT_CHANNEL, logoutUser);
+    createDispatchChannel(LOGIN_CHANNEL, loginUser);
+  }, [dispatch]);
```
Alessandro100, line 46: how come we use `[dispatch]` over `[]` if we want these channels to be init once?

web-app/src/app/router/Router.tsx
```diff
@@ -17,8 +17,34 @@ import TermsAndConditions from '../screens/TermsAndConditions';
...
+  const navigateTo = useNavigate();
+  const dispatch = useAppDispatch();
+  const logoutUser = (): void => {
+    dispatch(
+      logout({ redirectScreen: SIGN_OUT_TARGET, navigateTo, propagate: false }),
+    );
+  };
+
+  const loginUser = (): void => {
+    // Refresh the page to ensure the user is authenticated
```
Alessandro100, line 39: I would update `Refresh the page to ensure the user is authenticated on this session and other sessions (tabs)`

davidgamez: The refresh will only affect the current _window/session_

web-app/src/app/router/Router.tsx
```diff
@@ -17,8 +17,34 @@ import TermsAndConditions from '../screens/TermsAndConditions';
...
+    window.location.reload();
+  };
+
+  useEffect(() => {
+    createDispatchChannel(LOGOUT_CHANNEL, logoutUser);
+    createDispatchChannel(LOGIN_CHANNEL, loginUser);
```
Alessandro100, line 45: [question] how come we created a custom subscribe / dispatch system instead of using Redux reducer `dispatch` directly?

web-app/src/app/router/Router.tsx
davidgamez: The dispatch instance does not change between renders and guarantees that the app is in _ready_ state. I can switch it to _[]_, it have the same effect.

web-app/src/app/router/Router.tsx
davidgamez: The redux dispatch doesn't dispatch events across tabs; it's scoped to the current session. We are using redux-persist, but the loading of the state from the local storage is only applicable on redux state rehydration(when the app loads the first time or refreshes). I was looking into other redux libraries that sync states across tabs. However, those libraries make the tabs entirely in sync, making it tricky to have different "states" reflected on each tab. Exanple open different search results or feeds.

Alessandro100 (approved):
✅

web-app/src/app/router/Router.tsx
davidgamez: Fixed!

web-app/src/app/router/Router.tsx
davidgamez: I added documentation and renamed variable and function names to improve readability.

PR #545, opened 2024-07-09
fix: failing integration tests 

cka-y:
**Summary:**

- **Fixed Failing Integration Tests:**
  - Implemented an outer join for the `location` and `feed` entities to ensure proper data fetching.
  - Added `joinedload` for all entities accessed by the `from_orm` methods to optimize query performance and reduce additional database queries. For instance, querying `https://api-qa.mobilitydatabase.org/v1/gtfs_feeds` now takes approximately 1.6 seconds, down from 17 seconds.
- Removed the unnecessary test for the validity of the latest URL, as this validation is already handled within the data pipeline. Performing this test within the integration tests is redundant and not the most appropriate place for it.

[Example of successful execution](https://github.com/MobilityData/mobility-feed-api/actions/runs/9863354004/job/27236236214)

### DIFF
--- a/api/src/feeds/impl/feeds_api_impl.py
+++ b/api/src/feeds/impl/feeds_api_impl.py
@@ -1,8 +1,7 @@
 from datetime import datetime
 from typing import List, Union
 
-from sqlalchemy import select
-from sqlalchemy.orm import aliased, contains_eager, joinedload
+from sqlalchemy.orm import joinedload
 
 from database.database import Database
 from database_gen.sqlacodegen_models import (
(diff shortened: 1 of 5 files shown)


api/src/feeds/impl/feeds_api_impl.py
```diff
@@ -155,6 +161,21 @@ def get_gtfs_feeds(
...
+        gtfs_feed_query = (
+            gtfs_feed_query.outerjoin(Location, Feed.locations)
+            .outerjoin(latest_datasets, Gtfsfeed.gtfsdatasets)
+            .options(
+                contains_eager(Gtfsfeed.gtfsdatasets, alias=latest_datasets)
+                .joinedload(Gtfsdataset.validation_reports)
+                .joinedload(Validationreport.notices),
+                *BasicFeedImpl.get_joinedload_options(),
+            )
+        )
```
davidgamez, line 178: How can we benefit from the latest datasets query if the _GtfsRTFeedImpl.from_orm_ is still looping over all datasets?

cka-y: Indeed! There's no way to load the latest dataset exclusively so I removed this query.

davidgamez: We will hit a performance issue in the future as the datasets keep growing. We can revisit mapping the latest dataset as part of the entity. I remember having a few challenges with the previous SqlAchemy version; maybe the new version is easier to implement. Created an issue for it #548

davidgamez (approved):
LGTM!

PR #631, opened 2024-07-24
feat: create a cloud function to extract bounding boxes in batch

cka-y:
**Summary:**
Closes #590
This PR introduces a batch version of the cloud function responsible for extracting bounding boxes from datasets. This enhancement addresses the need for a more efficient processing method, allowing multiple datasets to be processed concurrently.

**Changes Made:**
- **Conversion to Pub/Sub Triggered Function:** The core logic of the function has been moved to a Pub/Sub triggered function. This change enables the batch process to initiate multiple dataset extractions without having to wait for each one to complete sequentially.
- **Retention of Eventarc Triggered Function:** The existing function triggered by Eventarc, which activates when a new dataset is uploaded to the GCP datasets bucket, remains unchanged.
- **New HTTP Triggered Batch Function:** A new HTTP triggered function has been introduced to identify all `latest` datasets lacking a bounding box. This function publishes messages to the Pub/Sub topic to trigger the extraction process.

**Expected Behavior:** 
Bounding boxes are extracted for latest datasets that are missing them.

**Testing Tips:**
1. **Set Up Database Tunnel:** Establish a database tunnel with the development environment locally.
2. **Simulate Missing Bounding Boxes:** Delete bounding boxes from a selected subset of datasets.
3. **Deploy Branch in Development:** Deploy the branch in the development environment.
4. **Run Batch Process:** Use the following curl command to trigger the batch function:
   ```bash
   curl -m 550 -X POST https://northamerica-northeast1-mobility-feeds-dev.cloudfunctions.net/extract-bounding-box-batch \
   -H "Authorization: bearer $(gcloud auth print-identity-token)" \
   -H "Content-Type: application/json"
   ```
5. **Verify Logs:** Check the logs of the `extract-bounding-box-batch` function to ensure that the bounding box extraction ran successfully for the datasets with deleted bounding boxes.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Run the unit tests with `./scripts/api-tests.sh` to make sure you didn't break anything
- [ ] Add or update any needed documentation to the repo
- [ ] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] Linked all relevant issues
- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)

Issue #618, opened 2024-07-22
Automate location generation


emmambd:
### Describe the problem

Defined in #564

### Proposed solution

**Methodology:**
A sample of 5 stops is extracted from the `stops.txt` file. If more than half of the sampled stops within a feed have different municipalities or subdivisions, these specific values are not considered in determining the feed's location and are set to `None`. However, if the feed includes multiple countries, the list of all countries is included in the feed's location information.

- To test that it works
    - I can input full country name and return a series of results 
    - I can input a location name without accents and it works

emmambd:
@emmambd to adjust UI design to incorporate multiple countries

cka-y:
Tasks:
1. **Modify the database entity** to support multiple languages.
2. **Add a link between the `Location` entity and the `GTFSDataset` entity**, as the location will now be extracted from the bounding box of the dataset.
3. **Modify/set the location in the datasets** by calling the Nominatim API in both the default and English languages. This can be done as part of the `extract_bb` cloud function, which could be renamed to `extract_location`.
4. **Modify the API response for the feed** to set the location as the latest dataset location if that dataset exists; otherwise, resort to the location information extracted from the `mobility-database-catalog`.
5. **Modify the materialized view utilized by the search endpoint** to fix issues related to searching for locations in English.

PR #635, opened 2024-07-25
fix: infinite retries of `extract_bounding_box_pubsub`

cka-y:
### Summary:
Part of #618 — This update implements necessary fixes for the bounding box extraction function, enabling future location extraction:
- **Error Handling**: Added missing try/catch blocks to prevent infinite retries.
- **Resource Allocation**: Increased memory and CPU limits because of insufficient memory problems.
- **Trace Logging**: Introduced logging for bounding box extraction, with updates to `DatasetTraceService` and the `extract_bb` module.

### Expected Behavior:
Enhancements in error handling, resource management, and visibility for bounding box extraction processes.

### Testing Tips:
- Follow testing procedures from PR #631.
- Ensure processes skip if execution limits exceed (`MAXIMUM_EXECUTIONS` default is `1`).
- Check the functionality and logging accuracy for bounding box extraction.

image

Testers are encouraged to explore beyond these tips to identify any potential issues.

### DIFF
--- a/functions-python/dataset_service/main.py
+++ b/functions-python/dataset_service/main.py
@@ -134,7 +134,7 @@ class DatasetTraceService:
             file_sha256_hash=entity.get("file_sha256_hash"),
             hosted_url=entity.get("hosted_url"),
             error_message=entity.get("error_message"),
-            pipeline_stage=PipelineStage(entity.get("pipeline_stage")),
+            pipeline_stage=PipelineStage(entity.get("pipeline_stage")) if entity.get("pipeline_stage") else None,
             dataset_id=entity.get("dataset_id"),
         )
 
--- a/functions-python/dataset_service/main.py
+++ b/functions-python/dataset_service/main.py
@@ -118,6 +118,7 @@ class DatasetTraceService:
         entity.update(asdict(dataset_trace))
         entity["trace_id"] = trace_id
         entity["status"] = dataset_trace.status.value
+        entity["pipeline_stage"] = dataset_trace.pipeline_stage.value
 
         return entity
 
@@ -133,6 +134,10 @@ class DatasetTraceService:
             file_sha256_hash=entity.get("file_sha256_hash"),
             hosted_url=entity.get("hosted_url"),
             error_message=entity.get("error_message"),
+            pipeline_stage=PipelineStage(entity["pipeline_stage"])
+            if "pipeline_stage" in entity
+            else None,
+            dataset_id=entity.get("dataset_id"),
         )
 
 
--- a/functions-python/dataset_service/main.py
+++ b/functions-python/dataset_service/main.py
@@ -134,7 +134,9 @@ class DatasetTraceService:
             file_sha256_hash=entity.get("file_sha256_hash"),
             hosted_url=entity.get("hosted_url"),
             error_message=entity.get("error_message"),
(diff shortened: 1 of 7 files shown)


functions-python/dataset_service/main.py
```diff
@@ -120,6 +134,10 @@ def _entity_to_dataset_trace(entity: datastore.Entity) -> DatasetTrace:
             file_sha256_hash=entity.get("file_sha256_hash"),
             hosted_url=entity.get("hosted_url"),
             error_message=entity.get("error_message"),
+            pipeline_stage=PipelineStage(entity["pipeline_stage"])
```
jcpitre, line 137: Please educate me: Why do we sometimes use .get() and sometimes the square brackets.

qcdyx: There you go. 
In Python, dictionaries provide two primary ways to access their values: using square brackets ([]) and the .get() method. Each has its own use cases and advantages.

Square Brackets ([])
Usage: value = dictionary[key]
Behavior: Raises a KeyError if the key is not found in the dictionary.
Use Case: Use square brackets when you are certain that the key exists in the dictionary. This is the most straightforward and fastest way to access a value.

.get() Method
Usage: value = dictionary.get(key, default_value)
Behavior: Returns None (or a specified default value) if the key is not found in the dictionary.
Use Case: Use .get() when you are not sure if the key exists in the dictionary and you want to avoid a KeyError. It is also useful when you want to provide a default value if the key is missing.

qcdyx (approved):
LGTM

functions-python/dataset_service/main.py
jcpitre: Thanks
So in the above code we are certain that `pipeline_stage` exists, but not `error_message` or `hosted_url` ?

cka-y: Good catch! it is indeed a mistake on my part -- fixing it

jcpitre: Thanks,
So in the above code we are sure `timestamp` exists for example, but not execution_id?

cka-y: exactly -- if you refer to the `DatasetTrace` class:
https://github.com/MobilityData/mobility-feed-api/pull/635/files/fc0c4cc5ffdefed5ab7dbd684e578cd87215fcbe#diff-30e7f65b1f1c389dc41778ebd8099c9e64edee06e5a0a1924fa6112d95f03727R49-R60, the `execution_id` is `Optional` whereas `timestamp` isn't. 

The `pipeline_stage` isn't optional, but your comment is still pertinent because since the field has been added after the datastore kind has been created, any historical entity will have an empty/`null` value for that field, so using `get` instead of the square brackets is safer for sure.

infra/functions-python/main.tf
```diff
@@ -267,7 +267,7 @@ resource "google_cloudfunctions2_function" "extract_bb_batch" {
       PUBSUB_TOPIC_NAME = google_pubsub_topic.dataset_updates.name
       PYTHONNODEBUGRANGES = 0
     }
-    available_memory = local.function_extract_bb_config.memory
+    available_memory = "1Gi"
```
davidgamez, line 270: Why not use the value from the config file?

cka-y: the value in the config file is higher, but the batch function entry point does not require as much memory. I also didn't feel it was necessary to introduce a new variable for this purpose.

davidgamez (approved):
LGTM

PR #660, opened 2024-08-06
fix: 649 search screen back and forward browsers actions reacting on typying

### LINKED ISSUE #649: Search screen back and forward browser's actions reacting on typying
### Describe the bug

The back and forward browser's history saves changes while the user types in the search box. 

### Steps/Code to Reproduce

- Go to feeds screen
- Start typing 
- Use the back and forward action and see the screen don't change only change the search parameter in the URL

### Expected Results

The history is only saved when the search action is dispatched.

### Actual Results

Browser history is saved per typed character.

### Screenshots

https://github.com/user-attachments/assets/97cf08d9-6be9-4f7e-8f1d-4900a96f33a2

### Files used

_No response_

### Additional notes

_No response_

qcdyx:
**Summary:**
Closes #649 
update the URL query params only when the search action is dispatched, rather than with each character typed.

**Expected behavior:** 
search for new york in feeds screen, click back navigation button in the browser, it goes back to the previous page instead of  removing the previous keystroke.

**Testing tips:**

### DIFF
--- a/web-app/src/app/screens/Feeds/index.tsx
+++ b/web-app/src/app/screens/Feeds/index.tsx
@@ -126,6 +126,13 @@ export default function Feed(): React.ReactElement {
     setSearchParams(newSearchParams);
   }, [activeSearch, activePagination]);
 
+  useEffect(() => {
+    const newQeury = searchParams.get('q') ?? '';
+    if (newQeury !== searchQuery) {
+      setSearchQuery(newQeury);
+    }
+  }, [searchParams]);
+
   const getSearchResultNumbers = (): string => {
     if (feedsData?.total !== undefined && feedsData?.total > 0) {
       const offset = getPaginationOffset(activePagination);
@@ -173,7 +180,7 @@ export default function Feed(): React.ReactElement {
                 sx={{
                   width: 'calc(100% - 100px)',
                 }}
-                defaultValue={searchQuery}
+                value={searchQuery}
                 placeholder={t('searchPlaceholder')}
                 onChange={(e) => {
                   setSearchQuery(e.target.value);
--- a/web-app/src/app/screens/Feeds/index.tsx
+++ b/web-app/src/app/screens/Feeds/index.tsx
@@ -91,7 +91,9 @@ export default function Feed(): React.ReactElement {
   };
 
   const handleSearch = (): void => {
-    setSearchParams({ q: searchQuery });
+    if (searchQuery !== '') {
+      setSearchParams({ q: searchQuery });
+    }
     const paginationOffset = getPaginationOffset();
     if (user !== undefined) {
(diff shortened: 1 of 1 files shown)


davidgamez:
The initial search query from the home page is not set in the search query input, but the subsequence text is properly kept in it. The attached screen capture shows this.

https://github.com/user-attachments/assets/4f40c322-aaa7-452d-bcf0-5c7ee6f9c775

Alessandro100:
If we do a search and go back it wont update the search results. This also happens with pagination.

Expected result: If we go back, the data on the page should reflect the current url

https://github.com/user-attachments/assets/e8691603-dc3b-44e9-923c-89e09e264e58

web-app/src/app/screens/Feeds/index.tsx
```diff
@@ -71,11 +71,11 @@ export default function Feed(): React.ReactElement {
   const [selectedFeedTypes, setSelectedFeedTypes] = useState(
     getInitialSelectedFeedTypes(searchParams),
   );
-  const [activeSearch, setActiveSearch] = useState('');
+  const [activeSearch, setActiveSearch] = useState(searchParams.get('q') ?? '');
+  const [searchQuery, setSearchQuery] = useState(activeSearch);
```
davidgamez, line 75: [suggestion]: Having two state variables with the same "meaning" can be confusing and make it hard to follow the code. If possible, I suggest using a single state variable.

qcdyx:
I forgot to mention that there's a bug where the result page doesn't change when the back navigation button is clicked. This issue also occurs on the main branch. I suggest creating a new issue to track this. @emmambd  I'm currently working on the URL change and using a single state variable. @Alessandro100 @davidgamez

web-app/src/app/screens/Feeds/index.tsx
qcdyx: I intentionally use a 'duplicate' searchQuery to address the original issue. In line 178, the onChange function only updates the searchQuery, not the activeSearch, which is responsible for triggering an actual search. That's why it removed a keystroke when the back navigation button is clicked.

davidgamez:
> I forgot to mention that there's a bug where the result page doesn't change when the back navigation button is clicked. This issue also occurs on the main branch. I suggest creating a new issue to track this. @emmambd I'm currently working on the URL change and using a single state variable. @Alessandro100 @davidgamez

I checked in QA and the pagination issue is present, let's keep it outside this PR as @qcdyx suggested.

qcdyx:
https://github.com/MobilityData/mobility-feed-api/issues/667 @emmambd

Alessandro100 (approved):
LGTM, intuitive way of triggering the search ✅

PR #661, opened 2024-08-07
feat: automate locations - modified API responses

cka-y:
**Summary:**

This PR updates the feed endpoints and the search endpoint to use the extracted locations and includes a new `country` field for each location within the `gtfs_feeds` response.

**Out of Scope:** 
- Filtering behaviour should use the english version of the locations in #662 (this is also needed to update the integration tests)

**Expected Behavior:**
- Each location in the response now includes a `country` field.
- All location fields are provided in English.

**Testing Tips:**
1. Access the UI link provided in the PR comments.
2. Navigate through the feeds.
3. Inspect the network tab response to verify the presence of the `country` field.
4. Ensure that all location fields are in English.


integration-tests/src/endpoints/gtfs_feeds.py
```diff
@@ -26,21 +26,21 @@ def test_gtfs_feeds(self):
...
-    #     )
-    #     for i, country_code in enumerate(country_codes):
-    #         self._test_filter_by_country_code(
-    #             country_code,
-    #             "v1/gtfs_feeds",
-    #             validate_location=True,
-    #             task_id=task_id,
-    #             index=f"{i + 1}/{len(country_codes)}",
-    #         )
+    def test_filter_by_country_code_gtfs(self):
```
cka-y, line 29: This test should still work as the country codes are the same

cka-y:
The integration tests are expected to fail as i have introduced a new view that is not in QA yet

liquibase/changes/feat_622.sql
```diff
@@ -0,0 +1,28 @@
+DROP VIEW IF EXISTS location_with_translations;
+CREATE VIEW location_with_translations AS
```
davidgamez, line 2: If we are creating a view for translations limited to _en_, let's reflect this on the name, something like `location_with_translations_en`.

docs/DatabaseCatalogAPI.yaml
```diff
@@ -677,6 +677,10 @@ components:
             For a list of valid codes [see here](https://unece.org/trade/uncefact/unlocode-country-subdivisions-iso-3166-2).
           type: string
           example: US
+        country:
+            description: The name of the country where the system is located.
```
davidgamez, line 681: [suggestion]: Add a note to clarify that this field only supports English translations for now.

davidgamez:
Let's ignore/comment on the integration test to avoid data updates in PROD.

docs/DatabaseCatalogAPI.yaml
cka-y: this would apply to each field in the location -- should i specify it for each?

davidgamez: You are right; except country_code, I think is a good idea to be explicit about it as the data in the catalog repo might be in a different language(with the exception of the feeds without datasets that they will have the same value)

cka-y:
> Let's ignore/comment on the integration test to avoid data updates in PROD.

The integration tests that might cause issues post merge are already commented

api/src/feeds/impl/feeds_api_impl.py
```diff
@@ -164,7 +183,7 @@ def get_gtfs_feeds(
         dataset_latitudes: str,
         dataset_longitudes: str,
         bounding_filter_method: str,
-    ) -> List[GtfsFeed]:
+    ) -> List[Gtfsfeed]:
```
davidgamez, line 186: This should be the API model, not the DB one.

davidgamez (approved):
👏

PR #680, opened 2024-08-14
Feat: 647 display multiple countries format

### LINKED ISSUE #647: UI change: Displaying multiple countries
### Describe the problem

Once #622 and #623 are complete, we should modify the UI to display multiple countries. 

### Proposed solution

- [ ] Search page: multiple countries should display as their full English name, comma separated, in the location column of the table, e.g "Germany, Poland"
- [ ] Individual feed page: full English names of countries should be used, comma separated, e.g "Germany, Poland, Austria" in the location section. 
- [ ] Include country flag in the search result and location field in the individual feed page. Should be emoji of country flag. 

Examples: 
- 🇨🇦 Canada, Ontario, Ottawa
- 🇨🇦 Canada | 🇺🇸 United States

_No response_

### Additional context

_No response_

Alessandro100:
closes #647 
**Summary:**

Format the new country data as shown in the screenshots

**Expected behavior:** 

Location data should be shown as: 🇨🇦 Canada | 🇺🇸 United States

**Testing tips:**

Go on the search page and see if the location data is shown as above. TIP: Spain is a good search for many countries

- [X] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [X] Include screenshot(s) showing how this pull request works and fixes the issue(s)
Screenshot 2024-08-14 at 14 24 37

Screenshot 2024-08-14 at 14 34 45

### DIFF
--- a/web-app/src/app/services/feeds/utils.spec.ts
+++ b/web-app/src/app/services/feeds/utils.spec.ts
@@ -1,0 +1,52 @@
+import { getLocationName, type FeedLocation } from './utils';
+
+const mockLocationMultiple: FeedLocation = [
+  {
+    country_code: 'US',
+    country: 'United States',
+    subdivision_name: 'California',
+    municipality: 'Los Angeles',
+  },
+  {
+    country_code: 'CA',
+    country: 'Canada',
+    subdivision_name: 'Ontario',
+    municipality: 'Toronto',
+  },
+];
+
+const mockLocationSingle: FeedLocation = [
+  {
+    country_code: 'US',
+    country: 'United States',
+    subdivision_name: 'California',
+    municipality: 'Los Angeles',
+  },
+];
+
+const mockLocationMultipleMinimal: FeedLocation = [
+  {
+    country_code: 'ES',
+    country: 'Spain',
+  },
+  {
+    country_code: 'CA',
+    country: 'Canada',
+    subdivision_name: 'Ontario',
+    municipality: 'Toronto',
+  },
+];
+
+describe('Feeds Utils', () => {
+  it('should format location names correctly', () => {
+    expect(getLocationName(mockLocationMultiple)).toBe(
+      '🇺🇸 United States, California, Los Angeles | 🇨🇦 Canada, Ontario, Toronto',
+    );
+    expect(getLocationName(mockLocationSingle)).toBe(
+      '🇺🇸 United States, California, Los Angeles',
+    );
+    expect(getLocationName(mockLocationMultipleMinimal)).toBe(
+      '🇪🇸 Spain | 🇨🇦 Canada, Ontario, Toronto',
+    );
+  });
+});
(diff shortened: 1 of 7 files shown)


cka-y:
Could we sort the locations by country name when multiple locations are present? This would ensure consistency between the search results view and the individual feed's page. 
Or maybe this should be done in the backend side. Thoughts? (cc: @davidgamez)

web-app/src/app/services/feeds/utils.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  {
+    country_code: 'CA',
+    country: 'Canada',
+    subdivision_name: 'Ontario',
+    municipality: 'Toronto',
+  },
+];
+
+describe('Feeds Utils', () => {
+  it('should format location names correctly', () => {
```
davidgamez, line 41: 👍

web-app/src/app/services/feeds/utils.ts
```diff
@@ -45,3 +48,32 @@ export type AllDatasetType =
   | paths['/v1/gtfs_feeds/{id}/datasets']['get']['responses'][200]['content']['application/json']
   | paths['/v1/datasets/gtfs/{id}']['get']['responses'][200]['content']['application/json']
   | undefined;
+
+export function getLocationName(locations: FeedLocation | undefined): string {
```
davidgamez, line 52: Why not use the Locations type provided by the API models? I find it confusing to use singular FeedLocation for an array representation. https://github.com/MobilityData/mobility-feed-api/blob/b8dea9e4a7e1444f4849b703cb54dde6e7670623/web-app/src/app/services/feeds/types.ts#L327

web-app/src/app/services/feeds/utils.ts
```diff
@@ -1,8 +1,11 @@
-import { type paths } from './types';
+import countryCodeEmoji from 'country-code-emoji';
+import { type paths, type components } from './types';
 
 export type AllFeedsType =
   paths['/v1/search']['get']['responses'][200]['content']['application/json'];
 
+export type FeedLocation = components['schemas']['Locations'];
```
davidgamez, line 7: The same comment re: using singular FeedLocation instead of API model.

web-app/src/app/services/feeds/utils.ts
Alessandro100: Maybe I'm not understanding it but `FeedLocation` is set from the auto generated code `components['schemas']['Locations'];` maybe I should call it `FeedLocations` since there could be many locations

davidgamez: > Maybe I'm not understanding it but FeedLocation is set from the auto generated code components['schemas']['Locations']

I understand; I would like to know why there is a need to create a new type pointing to the same definition. In the hypothetical future that we add Location(s) to a different entity like Provider(that doesn't exist yet), FeedLocation(s) will be a confusing name, or we will need to create another type.

> maybe I should call it FeedLocations since there could be many locations

Yes, FeedLocations is a better name.

davidgamez: Make sense, _EntityLocations_ is a great alias name. Thanks!

davidgamez (approved):
LGTM!

PR #681, opened 2024-08-15
Translations

fredericsimard:
**Summary:**

Translated the UI labels in 10 languages, did manual fixes on mistakes added by translator.


web-app/public/locales/es/account.json
```diff
@@ -0,0 +1,47 @@
...
+       "toggleVisibility": "Activar o desactivar la visibilidad del token de actualización",
+       "copy": "Copiar token de actualización",
+       "copyToClipboard": "Copiar token de actualización al portapapeles"
+    },
+    "accessToken": {
+       "title": "Cómo generar el token de acceso en su aplicación",
+       "description": {
+          "pt1": "Siga",
+          "cta": "nuestra documentación de OpenAPI",
+          "pt2": "para autenticar su cuenta con el token de actualización".
```
davidgamez, line 21: Invalid period at the end of the field value.
```suggestion
          "pt2": "para autenticar su cuenta con el token de actualización."
```

web-app/public/locales/fr/common.json
```diff
@@ -2,8 +2,8 @@
     "copyToClipboard": "Copier dans le presse-papiers",
     "copied": "Copié!",
     "name": "Nom",
-    "email": "E-mail",
+    "email": "Courriel",
```
davidgamez, line 5: Is this a valid translation, or will e-mail work?

davidgamez (commented):
This is great! As this repository is in active development, let's make sure that we create a development process that is easy to follow. It might be great if we introduce a way in the process to **override** text generated by the tool with _community_ fixed text. Before making this public, we should implement the search in multiple languages as @emmambd suggested. I also suggest adding a _disclaimer_ text pointing out the process to contribute to translations.

web-app/public/locales/fr/common.json
fredericsimard: Yeah, courriel is the word we use in French.

fredericsimard:
@davidgamez I can install the automation to do the translation when a PR is opened for these specific files and then added to the PR for someone to revise and change what needs to. It's untested in prod but should work. I was waiting for the new gtfs.org website to be online before installing it there, but it's ready.

davidgamez:
> @davidgamez I can install the automation to do the translation when a PR is opened for these specific files and then added to the PR for someone to revise and change what needs to. It's untested in prod but should work. I was waiting for the new gtfs.org website to be online before installing it there, but it's ready.

Let's connect to go over the details.

davidgamez (approved):

PR #701, opened 2024-08-20
fix: Set the country if missing from DB in the populate script

### LINKED ISSUE #694: Missing `country` value in location entities with no extracted location data
### Describe the bug

GTFS feeds with no extracted location data and their corresponding GTFS RT feeds are resulting in an empty `country` field for the associated `Location` entity in the database. In cases where a `country` value is present, it is because the extracted location matches an existing location (`location_id`), automatically setting the `country` value. However, this behavior is inconsistent across the database, leading to missing `country` values for many `Location` entities. This affects the search results when searching by country name.

jcpitre:
**Summary:**
Closes #694 

Added code to set the country in the populate script. This will work only for new entries
For old entries, added a post-processing operation in the populate script that looks at all the location table entries and adds the country if it's missing. The country is obtained from the country code already in the table entry.

**Expected behavior:** 

All locations should now have a country if the country code has a corresponding country

**Testing tips:**

Tested locally with a location table on which I set all countries to null. Test that the post processing sets the countries properly.
Tested also by deleting the DB and rerun the populate script. The countries should also be set properly as they are considered new entries.

### DIFF
--- a/api/src/scripts/populate_db_gtfs.py
+++ b/api/src/scripts/populate_db_gtfs.py
@@ -1,6 +1,7 @@
 import os
 from datetime import datetime
 
+import pycountry
 import pytz
 from sqlalchemy import text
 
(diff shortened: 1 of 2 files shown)


jcpitre:
For reference, there was 1351 locations without country in the QA DB.
Ran the db-update-qa.yml GH action.
Got this log:
`024-08-20 18:55:45,732 INFO GTFSDatabasePopulateHelper Had to set the country for 1350 locations`

Looked in the DB. There is still 1 location still with country = null,  but it's the one with all nulls (country_code=null, subdivision=null, etc)

Spot checked by using the QA API with items from the list @cka-y gave us (https://github.com/MobilityData/mobility-feed-api/issues/694#issuecomment-2296757321) and they have a proper country now.

PR #766, opened 2024-10-16
docs: Update README.md

emmambd:
I realized before I was about to update the catalogs repo that our info on the API itself is out of date! Updated.


davidgamez (approved):
⭐

PR #837, opened 2024-11-27
feat: Use psycopg2 for Connection Pooling and Implement Global Engine with Context Manager for Session Management

### LINKED ISSUE #826: Use a global SQLAlchemy engine for db connection pooling with psycopg2
Refactor the code to use a shared engine upon open and close session per request, avoiding redundant engine/pool creation. A global engine is safe to share across threads because SQLAlchemy's Engine object is designed to be thread-safe. Using a global engine ensures that all database interactions reuse the same pool of connections without creating redundant engines.

qcdyx:
**Summary:**

Closes #826 

- Updated the database connection setup to use psycopg2 for connection pooling.
- Refactored the code to use a global SQLAlchemy engine.
- Implemented a context manager for managing database sessions.
- Ensured session isolation for each request to maintain data integrity and prevent conflicts.

**Expected behavior:** 

- database connections are reused, reducing the overhead of creating new connections and improving performance.
- the context manager ensures that sessions are properly opened, committed, rolled back, and closed, providing robust transaction management.
- ensured that each request gets its own isolated session.

Load test shows that ramping to 5000 users at a rate of 50.00 per second uses 22 active connections - we have max 100 connections.
image

I'll work on a knowledge share sub-issue and explain all the changes in details.

**Testing tips:**


api/src/database/database.py
```diff
@@ -59,7 +75,7 @@ class Database:
 
     def __new__(cls, *args, **kwargs):
         if not isinstance(cls.instance, cls):
-            with lock:
+            with cls.lock:
```
davidgamez, line 78: [question]: Why the class lock instead of the file level lock?

api/src/database/database.py
```diff
@@ -77,82 +93,44 @@ def __init__(self, echo_sql=False):
...
-            self.engine = None
+            self.logger = logging.getLogger(__name__)
             self.connection_attempts = 0
-            self.SQLALCHEMY_DATABASE_URL = os.getenv("FEEDS_DATABASE_URL")
-            self.echo_sql = echo_sql
-            self.start_session()
+            database_url = os.getenv("FEEDS_DATABASE_URL")
+            if database_url is None:
+                raise Exception("Database URL not provided.")
+            self.engine = create_engine(database_url, echo=echo_sql, pool_size=10, max_overflow=0)
```
davidgamez, line 104: [suggestion]: have the pool size as an os variable, so we can play with the size without the need of re-package the python function

api/src/database/database.py
```diff
@@ -48,6 +44,26 @@ def configure_polymorphic_mappers():
     gbfsfeed_mapper.polymorphic_identity = Gbfsfeed.__tablename__.lower()
 
 
+def with_db_session(func):
```
davidgamez, line 47: 👍

api/src/scripts/populate_db_gtfs.py
```diff
@@ -236,40 +244,41 @@ def trigger_downstream_tasks(self):
...
+                self.logger.info("Refreshing MATERIALIZED FEED SEARCH VIEW - Started")
+                session.execute(text(f"REFRESH MATERIALIZED VIEW CONCURRENTLY {t_feedsearch.name}"))
+                self.logger.info("Refreshing MATERIALIZED FEED SEARCH VIEW - Completed")
+                session.commit()
+                self.logger.info("\n----- Database populated with sources.csv data. -----")
+                if trigger_downstream_tasks:
+                    self.trigger_downstream_tasks()
         except Exception as e:
             self.logger.error(f"\n------ Failed to populate the database with sources.csv: {e} -----\n")
-            self.db.session.rollback()
```
davidgamez, line 272: [question]: should we rollback here?

api/tests/test_utils/db_utils.py
```diff
@@ -204,13 +204,13 @@ def empty_database(db, url):
...
+                for table_name in tables_to_delete:
+                    table = Base.metadata.tables[table_name]
+                    delete_stmt = delete(table)
+                    session.execute(delete_stmt)
 
-            db.commit()
+                session.commit()
 
         except Exception as error:
-            db.session.rollback()
```
davidgamez, line 215: [question]: Similar to the previous comment, why not rollback?

api/src/scripts/populate_db_gtfs.py
```diff
@@ -18,6 +19,9 @@
 from scripts.load_dataset_on_create import publish_all
 from utils.data_utils import set_up_defaults
 
+if TYPE_CHECKING:
+    from sqlalchemy.orm import Session
```
davidgamez, line 23: [question]: Why do we need this check here? Is there any circular dependency between files/modules that we need to look at?

api/src/scripts/populate_db_gtfs.py
qcdyx: rollback will be called inside contextmanager [here](https://github.com/MobilityData/mobility-feed-api/blob/bae94e06f77ddc49bd8672b6275097d2e45ec4e1/api/src/database/database.py#L121-L122)
```Python
    @contextmanager
    def start_db_session(self):
        session = self.Session()
        try:
            yield session
            session.commit()
        except Exception:
            session.rollback()
            raise
        finally:
            session.close()
```

api/tests/test_utils/db_utils.py
qcdyx: same as above

api/src/scripts/populate_db_gtfs.py
qcdyx: There's no circular dependency identified. The `Session` is not actually used. I added it just to have a better type hint. https://gitkraken.dev/link/dnNjb2RlOi8vZWFtb2Rpby5naXRsZW5zL2xpbmsvci8xNmE2MTAyMDY4N2VmNWY4NDg4OTgwZWJiYTNlYzZhNGI1MzUwYjRiL2YvYXBpL3NyYy9zY3JpcHRzL3BvcHVsYXRlX2RiX2d0ZnMucHk%2FdXJsPWdpdCU0MGdpdGh1Yi5jb20lM0FNb2JpbGl0eURhdGElMkZtb2JpbGl0eS1mZWVkLWFwaS5naXQmbGluZXM9OTgtOTk%3D?origin=gitlens

api/src/database/database.py
qcdyx: This is a good question, there are two locks in this file, one class level and one file level. We can open a new issue to investigate it.

api/src/scripts/populate_db_gtfs.py
qcdyx: **Conditional Import:**
The [if TYPE_CHECKING:](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) block ensures that the [Session](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) import is only included during type checking and not at runtime.
This helps avoid unnecessary imports and keeps the runtime environment clean.

**Type Hinting:**
The session: 'Session' type hint in the function signature uses a forward reference (enclosed in quotes) to refer to the [Session](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) type.
This allows you to provide type hints without actually importing the [Session](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) class at runtime.

davidgamez: I had the understanding(maybe I'm wrong) that the type_checking is mainly used when there is a potentially circular import. This avoids the circular import, as the imported target is not used at runtime. PyCharm gives type hinting while on regular imports. I think you should only use the type_ckecking if needed; otherwise, why use it just for the Session? We can go over a more detailed conversation tomorrow morning.

api/src/scripts/populate_db_gtfs.py
davidgamez: I got it; thanks for the clarification.

api/src/database/database.py
davidgamez: To avoid confusion, we can delete the file level lock if it's not in use.

api/tests/test_utils/db_utils.py
```diff
@@ -204,13 +204,13 @@ def empty_database(db, url):
...
-                delete_stmt = delete(table)
-                db.session.execute(delete_stmt)
+            with db.start_db_session() as session:
+                for table_name in tables_to_delete:
+                    table = Base.metadata.tables[table_name]
+                    delete_stmt = delete(table)
+                    session.execute(delete_stmt)
 
-            db.commit()
+                session.commit()
```
jcpitre, line 213: Do we need the commit here? I thought the commit would be done in start_db_session. I guess an extra commit is not a big deal.

api/src/scripts/populate_db_gtfs.py
qcdyx: We agree to remove it in another PR.

api/src/database/database.py
qcdyx: Done

api/src/database/database.py
qcdyx: Done.

api/tests/test_utils/db_utils.py
qcdyx: removed.

davidgamez (approved):
Amazing improvement!

Issue #898, opened 2025-01-31
Export CSV function exceeds memory limit


davidgamez:
### Describe the bug

The execution of the export CSV function exceeds the assigned memory limit in PROD.

Related to: #888 

### Steps/Code to Reproduce

- Execute the `export.csv` function in the prod

### Expected Results

- Function completes and generates the CSV file at https://files.mobilitydatabase.org/sources_v2.csv

### Actual Results

- The function fails with error in the logs
```
Memory limit of 1024 MiB exceeded with 1028 MiB used. Consider increasing the memory limit.....
```

### Additional notes

I increased the memory to 2GB and got the function running. However, this solution is not permanent, as the feed and dataset counts will continue to grow exponentially.

PR #888, opened 2025-01-21
779 generate new sourcescsv from postgresql database2

jcpitre:
**Summary:**

Summarize the changes in the pull request including how it relates to any issues (include the #number, or link them).

**Expected behavior:** 

Explain and/or show screenshots for how you expect the pull request to work in your testing (in case other devices exhibit different behavior).

**Testing tips:**

Provide tips, procedures and sample files on how to test the feature.
Testers are invited to follow the tips AND to try anything they deem relevant outside the bounds of the testing tips. 

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Run the unit tests with `./scripts/api-tests.sh` to make sure you didn't break anything
- [ ] Add or update any needed documentation to the repo
- [ ] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] Linked all relevant issues
- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)

PR #900, opened 2025-02-02
feat: reduce memory usage for export_csv

sylvansson:
**Summary:**

This PR attempts to help reduce memory usage for the export CSV function (#898) by writing to the CSV file incrementally instead of accumulating data in a DataCollector instance and writing at the very end. 

I don't know how much data is read from the production database with the current queries so I can't really quantify the impact of my changes, but this is definitely not a long-term solution. We could probably reduce memory overhead much further by reading feed data in smaller batches using something like [`yield_per`](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.yield_per), but I preferred not touching that code because I'm not familiar with SQLAlchemy. I could address that in a subsequent PR.

**Expected behavior:** 

The export CSV function should use less memory but produce the same output.

**Testing tips:**

Based on #888, we should able to run the api-dev workflow and test against the Dev environment, but as far as I can't tell I (understandably) don't have the ability to do that :-)

### DIFF
--- a/functions-python/export_csv/src/main.py
+++ b/functions-python/export_csv/src/main.py
@@ -37,7 +37,6 @@ from shared.helpers.database import Database
 
 load_dotenv()
 csv_default_file_path = "./output.csv"
-csv_file_path = csv_default_file_path
 
 
 class DataCollector:
@@ -71,12 +70,13 @@ class DataCollector:
 
 @functions_framework.http
 def export_and_upload_csv(request=None):
-    response = export_csv()
+    csv_file_path = csv_default_file_path
+    response = export_csv(csv_file_path)
     upload_file_to_storage(csv_file_path, "sources_v2.csv")
     return response
 
 
-def export_csv():
+def export_csv(csv_file_path: str):
     """
     HTTP Function entry point Reads the DB and outputs a csv file with feeds data.
     This function requires the following environment variables to be set:
@@ -138,7 +138,6 @@ def collect_data() -> DataCollector:
     except Exception as error:
         logging.error(f"Error retrieving feeds: {error}")
         raise Exception(f"Error retrieving feeds: {error}")
-    data_collector.write_csv_to_file(csv_file_path)
     return data_collector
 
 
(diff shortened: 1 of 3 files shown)


functions-python/export_csv/src/main.py
```diff
@@ -30,53 +30,48 @@
...
+    "urls.authentication_info",
+    "urls.api_key_parameter_name",
+    "urls.latest",
+    "urls.license",
+    "location.bounding_box.minimum_latitude",
+    "location.bounding_box.maximum_latitude",
+    "location.bounding_box.minimum_longitude",
+    "location.bounding_box.maximum_longitude",
+    "location.bounding_box.extracted_on",
+    "status",
```
sylvansson, line 66: @jcpitre I noticed we don't set the status for realtime feeds, is that intentional?

davidgamez:
> Based on https://github.com/MobilityData/mobility-feed-api/pull/888, we should able to run the api-dev workflow and test against the Dev environment, but as far as I can tell I (understandably) don't have the ability to do that :-)

Yes, that specific requirement in the issue is for our internal team. For security and cost reasons, we don't share our environments' DB. 
I did test this branch with QA/DEV and PROD data, and it looks like memory usage was reduced by ~45%. Great work!

davidgamez:
> I don't know how much data is read from the production database with the current queries so I can't really quantify the impact of my changes, but this is definitely not a long-term solution. We could probably reduce memory overhead much further by reading feed data in smaller batches using something like [yield_per](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.yield_per), but I preferred not touching that code because I'm not familiar with SQLAlchemy. I could address that in a subsequent PR.

Aside from this PR improvement (awesome work 🥳  ), the main driver for memory increase is the fact that we are downloading datasets twice a week. This means that if the feed total stays the same(that is actually increasing with newcomers every week/month), the data will still grow with the new datasets. This is why I will keep the original issue open until we address this cause. The next round of optimization should look into the query and get only the latest dataset from each feed(or a limited number of datasets ordered by date).

functions-python/export_csv/src/main.py
davidgamez: This was not intentional, the realtime feeds also should have the `status` column populated.

davidgamez: I added a specific issue to address this, as we are missing a few other column values. #906

sylvansson:
Hi @davidgamez, thanks for the review. I'll fix the coverage % tonight. I'm also more than happy to tackle the other optimizations you've mentioned once this has gone out :-)

davidgamez (approved):
LGTM!

PR #907, opened 2025-02-08
fix: fetch feeds in batches for export_csv

sylvansson:
**Summary:**

This PR updates the export CSV function to fetch data in batches to reduce its memory usage (#898). My first approach used [`yield_per`](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.yield_per) which resulted in a large number of queries due to eager loading not being allowed (see fe236b6bef47fc0f7a17530e641fe3bca7dceff4). Instead we get a full list of feeds ids and then fetch data in batches using the existing queries.

**Expected behavior:** 

The export CSV function should use less memory but produce the same output.

**Testing tips:**

Test against all three environments and check the memory usage.


api/src/shared/common/iter_utils.py
```diff
@@ -0,0 +1,13 @@
+from itertools import islice
```
sylvansson, line 1: Initially wanted to put this file under api/src/utils but ran into import issues when running the tests so I chose the path of least resistance. Happy to move it.

functions-python/export_csv/src/main.py
```diff
@@ -114,33 +114,19 @@ def fetch_feeds() -> Iterator[Dict]:
...
+            for feed in get_all_gtfs_feeds(session, include_wip=False):
                 yield get_feed_csv_data(feed)
+                feed_count += 1
 
-            logging.info(f"Processed {len(gtfs_feeds)} GTFS feeds.")
+            logging.info(f"Processed {feed_count} GTFS feeds.")
 
-            for feed in gtfs_rt_feeds:
+            rt_feed_count = 0
+            for feed in get_all_gtfs_rt_feeds(session, include_wip=True):
```
jcpitre, line 125: Why include_wip=True?

sylvansson: Did some testing and forgot to set it back to false. Good catch, thank you :-)

jcpitre:
I did some testing on PROD. Memory usage went from 596MB to 112MB. Well done!

jcpitre (approved):
Solid contribution. Thanks!

jcpitre:
Ran the integration tests by hand. They passed.

PR #920, opened 2025-02-20
Feat: seo improvements v3 + feed detail performance

Alessandro100:
**Summary:**

General
- Assures the detailed description for feed detail page is set
- Updated sitemap
- Refactoring + fixing console errors

Feed Detail Page
- Added 'gtfs' / 'gtfs_rt' to URL
`feeds/gtfs` will re route to `feeds?gtfs=true` same with RT
- Optimized the way we get the feed detail data
old: 3 sequential API calls
now: 2 calls happening in parallel
- Removed duplicate GTFS RT data from display
- Prevent validation reports from being indexed

**Expected behavior:** 

With the feed detail page you can access the page either like
[old] `/feeds/mdb-2357`
[new] `/feeds/gtfs/mdb-2357` (faster)
same with `gtfs_rt`

**Testing tips:**

Navigate feed detail pages manually or through the feeds search page and assure that all the data is correct


cka-y:
When i visit `feeds/mdb-2347` I don't get redirected to `feeds/gtfs/mdb-2347`, which is the page i'm directed to when i select `Pace Bus` from the search screen. Is this expected behaviour?

web-app/src/app/screens/Feed/Feed.functions.tsx
```diff
@@ -0,0 +1,64 @@
...
+    feedName,
+  );
+  if (
+    sortedProviders.length === 0 &&
+    (feedName === undefined || feedName === '')
+  ) {
+    return '';
+  }
+  const dataTypeVerbose =
+    dataType === 'gtfs' ? t('common:gtfsSchedule') : t('common:gtfsRealtime');
```
davidgamez, line 46: [suggestion]: We must resolve this in a couple of places using the translated name from the data_type. Replacing the logic with a convention on the translation key would be easier. Examples, `common:dataType_gtfs` or `common:dataTypeGtfs`(this will require some process of the datatype from snake_case to camel_case). This can simplify the future introduction of GBFS data type.

web-app/src/app/store/store.ts
```diff
@@ -31,6 +31,7 @@ const sagaMiddleware = createSagaMiddleware();
 
 export const store = configureStore({
   reducer: persistedReducer,
+  devTools: process.env.NODE_ENV !== 'production',
```
davidgamez, line 34: [question]: are we setting NODE_ENV=production on release?

Alessandro100:
> When i visit `feeds/mdb-2347` I don't get redirected to `feeds/gtfs/mdb-2347`, which is the page i'm directed to when i select `Pace Bus` from the search screen. Is this expected behaviour?

yes this is expected behaviour, from `feeds/mdb-2347` we don't know if it's gtfs or gtfs_rt before making a call. We could make a redirect, but I don't think it impacts the user experience. Having the url without 'gtfs' 'gtfs_rt' is mainly for us

web-app/src/app/store/store.ts
Alessandro100: I didn't double check, but I noticed we used it already in our code in `web-app/src/app/screens/Account.tsx`

qcdyx (approved):
Thanks for the mob review, great work, LGTM!

web-app/src/app/screens/Feed/Feed.functions.tsx
davidgamez: This can be addressed in a follow-up PR.

davidgamez (approved):
LGTM!

Issue #953, opened 2025-03-06
Update Feed Status: Log Difference


Alessandro100:
In the python function `update_feed_status` it returns the number of feeds updated but not the number of feeds that actually changed.

We should log the difference in 'Active' 'Inactive' and 'Future' statuses.

emmambd:
Omg I would love to know this on a regular cadence

sylvansson:
Hi @emmambd, I'd be happy to pick this up unless Alessandro wants to. I could have a PR in the next few days.

sylvansson:
Great :-) Did you have a format in mind? I don't know how and where this gets logged but we could either squeeze everything on a single log line or pretty print:

```
status    before    after
--------  --------  -------
active    832       971
future    173       68
...       ...       ...
```

davidgamez:
> Great :-) Did you have a format in mind? I don't know how and where this gets logged but we could either squeeze everything on a single log line or pretty print:
> 
> ```
> status    before    after
> --------  --------  -------
> active    832       971
> future    173       68
> ...       ...       ...
> ```

Hi @sylvansson, it is preferable that logs be output in a structured format. We don't have consistency across the functions, but JSON-like is always preferable. This will help us in the future to standardize the consumption of the logs, as they are available in the cloud(GCP)

sylvansson:
That makes sense. It sounds like I should return an object instead of a string [here](https://github.com/MobilityData/mobility-feed-api/blob/d4bfb20107b4201460a0c4c72f086d4f2ebbd921/functions-python/update_feed_status/src/main.py#L92) then?

davidgamez:
> That makes sense. It sounds like I should return an object instead of a string [here](https://github.com/MobilityData/mobility-feed-api/blob/d4bfb20107b4201460a0c4c72f086d4f2ebbd921/functions-python/update_feed_status/src/main.py#L92) then?

That's correct!

davidgamez:
Implemented in #992

PR #992, opened 2025-03-15
feat: update update_feed_status to return status diff counts

sylvansson:
**Summary:**

This PR updates the update_feed_status function to return a dict with the number of feeds that have transitioned to each status (#953).

**Expected behavior:** 

Instead of returning this, where N is the number of rows that matched the filters, not necessarily the number of rows that transitioned to a different status:

> Script executed successfully. \<N\> feeds updated

We should get this:

```json
{
  "'inactive'::status": 19,
  "'future'::status": 87,
  "'active'::status": 153,
}
```

(We could also just use "inactive", "future", etc. as keys if we prefer that.) 

**Testing tips:**

This PR is still a draft because I haven't touched the tests yet. I see they use a mock session ([ref](https://github.com/MobilityData/mobility-feed-api/blob/6e8b3845a1e99610337773971f92902d7fc33dbe/functions-python/update_feed_status/tests/test_update_feed_status_main.py#L9)) but I think it'd be valuable to hit an actual database; I'd like to validate my changes before tackling this though ⚠️ 

### DIFF
--- a/functions-python/update_feed_status/src/main.py
+++ b/functions-python/update_feed_status/src/main.py
@@ -1,10 +1,11 @@
 import logging
 import os
 import functions_framework
+from collections import defaultdict
 from datetime import date
 from shared.helpers.logger import Logger
 from shared.helpers.database import Database
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Iterator, NamedTuple
 from sqlalchemy import case, text
 from shared.database_gen.sqlacodegen_models import Gtfsdataset, Feed
 
@@ -12,6 +13,48 @@ if TYPE_CHECKING:
     from sqlalchemy.orm import Session
 
 logging.basicConfig(level=logging.INFO)
+
+
+FEED_FILTERS = [
+    Feed.status != text("'deprecated'::status"),
+    Feed.status != text("'development'::status"),
+]
+
+
+class PartialFeed(NamedTuple):
+    """
+    Subset of the Feed entity with only the fields queried in `fetch_feeds`.
+    """
+    id: int
+    status: str
+
+
+def fetch_feeds(session: "Session") -> Iterator[PartialFeed]:
+    query = (
+        session
+        # When adding or removing fields here, `PartialFeed` should be updated to
+        # match, for type safety.
+        .query(Feed.id, Feed.status)
+        .filter(**FEED_FILTERS)
+        .yield_per(500)
+    )
+    for feed in query:
+        yield PartialFeed(id=feed.id, status=feed.status)
+
+
+def get_diff_counts(
+    feed_ids_by_before_status: defaultdict[str, set[int]],
+    session: "Session",
+) -> defaultdict[str, int]:
+    """
+    Get the number of feeds that have transitioned to each status. If a status
+    doesn't have new feeds, it's excluded from the response.
+    """
+    count_per_new_status: defaultdict[str, int] = defaultdict(int)
+    for feed in fetch_feeds(session):
+        if feed.id not in feed_ids_by_before_status[feed.status]:
+            count_per_new_status[feed.status] += 1
+    return count_per_new_status
 
 
 #  query to update the status of the feeds based on the service date range of the latest dataset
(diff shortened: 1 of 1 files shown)


functions-python/update_feed_status/tests/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+        "16",  # development
+        "25",
+    ]:
+        session.add(make_dataset(_id, True, past_date, future_date))
+
+    # -> future
+    for _id in [
+        "10",
+    ]:
+        session.add(make_dataset(_id, False, past_date, future_date))
```
sylvansson, line 81: @davidgamez @Alessandro100 Should a feed transition to "future" even if it has an active but non-latest dataset? The current logic allows it ([ref](https://github.com/MobilityData/mobility-feed-api/blob/6e8b3845a1e99610337773971f92902d7fc33dbe/functions-python/update_feed_status/src/main.py#L41)) but I don't know if it can happen.

functions-python/update_feed_status/src/main.py
```diff
@@ -14,6 +15,46 @@
...
+            Feed.status != text("'deprecated'::status"),
+            Feed.status != text("'development'::status"),
+        )
+        .yield_per(500)
+    )
+    for feed in query:
+        yield PartialFeed(id=feed.id, status=feed.status)
+
+
+def get_diff_counts(
```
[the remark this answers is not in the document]
sylvansson: That was my original approach. The issue with it is that if an active feed becomes inactive and an inactive feed becomes active, the counts won't change.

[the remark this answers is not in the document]
sylvansson: I wouldn't expect this approach to be memory hungry, but if you prefer, we could update feeds in batches and compute the diff incrementally.

functions-python/update_feed_status/tests/conftest.py
sylvansson: I'm not sure I follow, sorry. I'm basically wondering if we should set the status to "future" if a feed has two datasets: one that is current, and another (the latest) that starts in the future.

Alessandro100: The Feed status is based on service date range found in dataset. A feed could have many datasets, but only 1 latest dataset. The latest dataset == current (most up to date), so if the latest dataset has a service date range that starts in the future, then it will update the Feed status to future

sylvansson: Got it, thank you both for clarifying!

functions-python/update_feed_status/src/main.py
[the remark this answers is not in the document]
sylvansson: Unfortunately we can't just use `updated_count` because PG counts a row even if it didn't actually change, but we can work around that by filtering out feeds that already have the target status.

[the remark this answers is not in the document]
davidgamez: > Unfortunately we can't just use `updated_count` because PG counts a row even if it didn't actually change, but we can work around that by filtering out feeds that already have the target status.

That makes sense; the filters must be changed to get the correct count result.

functions-python/update_feed_status/tests/conftest.py
```diff
@@ -0,0 +1,89 @@
+#
```
sylvansson, line 1: Out of curiosity, do we actually need to include this in individual files?

functions-python/update_feed_status/src/main.py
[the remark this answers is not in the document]
sylvansson: This has been addressed in 008e90d.

functions-python/update_feed_status/tests/conftest.py
davidgamez: It's nice to have, many of the files don't have it

functions-python/update_feed_status/src/main.py
```diff
@@ -33,32 +33,40 @@ def update_feed_statuses_query(session: "Session"):
...
+        diff_counts: dict[str, int] = {}
+
+        for service_date_conditions, status in status_conditions:
+            diff_counts[status] = (
+                session.query(Feed)
+                .filter(
+                    Feed.id == latest_dataset_subq.c.feed_id,
+                    Feed.status != text("'deprecated'::status"),
+                    Feed.status != text("'development'::status"),
+                    # We filter out feeds that already have the status so that the
```
davidgamez, line 62: 👍

davidgamez (approved):
Very clean implementation. Thanks for your time and contribution!

PR #1053, opened 2025-03-26
fix: Now empty is_official value in CSV will not change the DB during populate

### LINKED ISSUE #1050: Empty is_officlal field in sources.csv will put false in DB and possibly overwrite the value already there
### Describe the bug

The code currently unilaterally sets the DB value to True if the csv field is True, False if the csv field is false.
But it will consider an empty csv field as false and then overwrite the DB value.
Since currently there is no official flag set in the json files of the catalog (i.e. all is_official fields in csv are empty)  all DB values will be set to False.

jcpitre:
**Summary:**
closes #1050 
When populating the DB from the CSV file, we erroneously took an empty is_official cell for a False. 

**Expected behavior:** 
Now an empty is_official cell of the CSV means do-not-change the value in the DB.
If the CSV contains TRUE or FALSE, the DB value will be updated accordingly.

**Testing tips:**
Tested manually by populating with 2 CSV files. The 1st file just set some TRUE and some FALSE values, and other were left empty.
The 2nd CSV file changed:

- one is_official TRUE to FALSE -> result: value False in DB
- one is_official FALSE to TRUE -> result: valueTrue in DB
- one is_official TRUE to Empty -> result: value stayed True in DB
- one is_official FALSE to Empty -> result: value stayed False in DB
For good measure I had one is_official in the CSV set to a random string, and it had no effect to the value in the DB

I still need to add automatic tests to support that.

### DIFF
--- a/api/src/scripts/populate_db.py
+++ b/api/src/scripts/populate_db.py
@@ -80,9 +80,28 @@ class DatabasePopulateHelper:
         """
         Get a safe value from the row
         """
-        if not row[column_name] or pandas.isna(row[column_name]) or f"{row[column_name]}".strip() == "":
-            return default_value if default_value is not None else None
-        return f"{row[column_name]}".strip()
+        value = row[column_name]
+        if not value or pandas.isna(value) or f"{value}".strip() == "":
+            return default_value
+        return f"{value}".strip()
+
+    @staticmethod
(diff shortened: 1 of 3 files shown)


qcdyx (approved):
LGTM

PR #1134, opened 2025-04-23
fix: sqlalchemy sync bugs + batch fill

### LINKED ISSUE #1072: Heatmap not displayed due to missing bounding box and reverse geolocation errors
### Describe the bug

Feeds like `mdb-503` have a `geolocation.geojson` file generated (dated March 12th), but it does not contain a bounding box. Consequently, the heatmap is not displayed, even though the bounding box is supposed to be generated before the heatmap. This issue persisted despite the latest dataset download on March 19th. 

It appears that the bounding box might be overwritten or deleted during the location extraction process. Additionally, the location was not extracted for the latest dataset, possibly because the stops did not change. However, there are numerous errors in the logs of the `reverse_geolocation_processor` indicating:
- Duplication of primary keys
- Blank values in required columns within an entity (potentially due to location changes)

### Steps/Code to Reproduce

1. Check the `geolocation.geojson` file for `mdb-503` and verify the absence of a bounding box.
2. Run the location extraction process on the latest dataset (downloaded on March 19th).
3. Inspect the logs of the `reverse_geolocation_processor` for errors related to primary key duplication and blank values in required columns.

### Expected Results

The latest dataset should have a bounding box, allowing the heatmap to be displayed and the location extraction should run without errors.

### Actual Results

The heatmap is not displayed. General logs show errors related to:
- Primary key duplication
- Blank required columns in entities, potentially due to location changes

Example of a feed wrong from the first dataset entry: https://mobilitydatabase.org/feeds/gtfs/mdb-2444

### Screenshots

_No response_

### Files used

_No response_

### Additional notes

- Verify whether the bounding box is overwritten or deleted during the location extraction process.
- Investigate why the location was not extracted for the latest dataset. This could be due to unchanged stops or errors within the `reverse_geolocation_processor` (expected behaviour).

cka-y:
###  **Summary**
Closes #1072.

This PR addresses an issue where certain feeds (e.g. [mdb-2444](https://mobilitydatabase.org/feeds/gtfs/mdb-2444)) were not processed correctly because their dataset zip file was uploaded to GCS **before** the corresponding database entity was created. Since the reverse geolocation process is triggered by the GCS event, it would attempt to fetch the dataset and fail when the DB record wasn’t available yet.

This behavior is similar to a previously resolved bug that was preventing validation reports from being processed. As the core logic already handles that case, no additional fix was necessary beyond ensuring proper sync timing.

Additionally, minor SQLAlchemy relationship sync issues were identified and resolved as part of this PR to prevent cascade-related errors and duplicate key issues.

A new optional query parameter `include_only_unprocessed` was added to the batch reverse geolocation endpoint. It defaults to `true` and allows targeting only feeds that haven't yet been processed — avoiding unnecessary re-processing of already handled feeds.

### **Expected Behavior**

- Reverse geolocation no longer fails when a feed’s dataset was added to GCS before the DB entity existed.
- The batch fill process can now skip feeds that already have location data.
- The optional query param can be toggled via `include_only_unprocessed=false` if you want to force reprocessing.

### DIFF
--- a/functions-python/reverse_geolocation/src/reverse_geolocation_processor.py
+++ b/functions-python/reverse_geolocation/src/reverse_geolocation_processor.py
@@ -215,6 +215,7 @@ def extract_location_aggregate(
     stop = Feedlocationgrouppoint(feed_id=feed_id, geometry=stop_point)
     stop.group = group
     db_session.add(stop)
+    db_session.flush()
     logging.info(
         f"Point {stop_point} matched to {', '.join([g.name for g in geopolygons])}"
     )
@@ -359,11 +360,13 @@ def extract_location_aggregates(
         )
         for location_group in location_aggregates.values()
     ]
-    gtfs_feed.feedosmlocationgroups = osm_location_groups
+    gtfs_feed.feedosmlocationgroups.clear()
+    gtfs_feed.feedosmlocationgroups.extend(osm_location_groups)
+
     for gtfs_rt_feed in gtfs_feed.gtfs_rt_feeds:
         logging.info(f"Updating GTFS-RT feed with stable ID {gtfs_rt_feed.stable_id}")
         gtfs_rt_feed.feedosmlocationgroups.clear()
-        gtfs_rt_feed.feedosmlocationgroups = osm_location_groups
+        gtfs_rt_feed.feedosmlocationgroups.extend(osm_location_groups)
 
     feed_locations = []
     for location_aggregate in location_aggregates.values():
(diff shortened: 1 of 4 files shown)


functions-python/reverse_geolocation/src/reverse_geolocation_processor.py
```diff
@@ -215,6 +215,7 @@ def extract_location_aggregate(
     stop = Feedlocationgrouppoint(feed_id=feed_id, geometry=stop_point)
     stop.group = group
     db_session.add(stop)
+    db_session.flush()
```
davidgamez, line 218: 👍

davidgamez (approved):
LGTM!

PR #1148, opened 2025-04-28
feat: implemented OR filter for feature search

### LINKED ISSUE #1105: Implement OR filter for feature search
### Describe the problem

Implement OR filter for feature search

### Proposed solution

See above

_No response_

### Additional context

_No response_

qcdyx:
**Summary:**

Closed #1105 
**Expected behavior:** 

image

**Testing tips:**

### DIFF
--- a/api/src/feeds/impl/models/search_feed_item_result_impl.py
+++ b/api/src/feeds/impl/models/search_feed_item_result_impl.py
@@ -42,11 +42,6 @@ class SearchFeedItemResultImpl(SearchFeedItemResult):
             ),
             redirects=feed_search_row.redirect_ids,
             locations=cls.resolve_locations(feed_search_row.locations),
-            features=(
-                sorted([feature for feature in feed_search_row.latest_dataset_features])
-                if feed_search_row.latest_dataset_features
-                else []
-            ),
             latest_dataset=(
                 LatestDataset(
                     id=feed_search_row.latest_dataset_id,
--- a/api/src/feeds/impl/models/search_feed_item_result_impl.py
+++ b/api/src/feeds/impl/models/search_feed_item_result_impl.py
@@ -33,37 +33,46 @@ class SearchFeedItemResultImpl(SearchFeedItemResult):
             feed_contact_email=feed_search_row.feed_contact_email,
             source_info=SourceInfo(
                 producer_url=feed_search_row.producer_url,
-                authentication_type=int(feed_search_row.authentication_type)
-                if feed_search_row.authentication_type
-                else None,
+                authentication_type=(
+                    int(feed_search_row.authentication_type) if feed_search_row.authentication_type else None
+                ),
                 authentication_info_url=feed_search_row.authentication_info_url,
                 api_key_parameter_name=feed_search_row.api_key_parameter_name,
                 license_url=feed_search_row.license_url,
             ),
             redirects=feed_search_row.redirect_ids,
             locations=cls.resolve_locations(feed_search_row.locations),
-            latest_dataset=LatestDataset(
-                id=feed_search_row.latest_dataset_id,
-                hosted_url=feed_search_row.latest_dataset_hosted_url,
-                downloaded_at=feed_search_row.latest_dataset_downloaded_at,
-                hash=feed_search_row.latest_dataset_hash,
-                service_date_range_start=feed_search_row.latest_dataset_service_date_range_start,
-                service_date_range_end=feed_search_row.latest_dataset_service_date_range_end,
-                agency_timezone=feed_search_row.latest_dataset_agency_timezone,
-                validation_report=LatestDatasetValidationReport(
-                    total_error=feed_search_row.latest_total_error,
-                    total_warning=feed_search_row.latest_total_warning,
-                    total_info=feed_search_row.latest_total_info,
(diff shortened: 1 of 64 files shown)


docs/DatabaseCatalogAPI.yaml
```diff
@@ -1120,6 +1127,55 @@ components:
...
+      in: query
+      description: Filter feeds by their GTFS features. [GTFS features definitions defined here](https://gtfs.org/getting-started/features/overview)
+      required: false
+      style: form
+      explode: false
+      schema:
+        type: array
+        items:
+          type: string
+          enum: # Copied from web-app/src/app/utils/consts.tsx
```
davidgamez, line 1141: I would set this type to just string, not enum. The reason is that this list is driven by the GTFS-Validator. Having a strict list will add complexity to deploying new versions of the validator, as it will require deployment synchronization. We can definitely add these values as a comment like: 
```
schema:
        type: array
        items:
          type: string
          description: |
             Allowed options: Text-To-Speech, Stops Wheelchair Accessibility, Trips Wheelchair Accessibility,......
             Other values are also accepted.
```

api/src/feeds/impl/models/search_feed_item_result_impl.py
```diff
@@ -33,37 +33,46 @@ def from_orm_gtfs(cls, feed_search_row: t_feedsearch):
...
-                    if feed_search_row.latest_dataset_features
-                    else [],
-                ),
-            )
-            if feed_search_row.latest_dataset_id
-            else None,
+            features=(
+                sorted([feature for feature in feed_search_row.latest_dataset_features])
+                if feed_search_row.latest_dataset_features
+                else []
```
davidgamez, line 48: Features are not part of the feed, but are only part of the latest dataset validation report. I think this is not necessary

docs/DatabaseCatalogAPI.yaml
```diff
@@ -4,7 +4,7 @@ info:
   title: Mobility Database Catalog
   description: |
     API for the Mobility Database Catalog. See [https://mobilitydatabase.org/](https://mobilitydatabase.org/).
-
```
davidgamez, line 7: This file contains too many changes due to formatting, making it very hard to review. Please only commit changes related to the PR.

web-app/src/app/screens/Feeds/index.tsx
```diff
@@ -124,7 +124,8 @@ export default function Feed(): React.ReactElement {
             is_official: isOfficialFeedSearch || undefined,
             // Fixed status values for now, until a status filter is implemented
             // Filtering out deprecated feeds
-            status: ['active', 'inactive', 'development', 'future'],
```
davidgamez, line 127: This looks like a merge conflict that got overridden. Please add back the `future` status.

docs/DatabaseCatalogAPI.yaml
```diff
@@ -1120,6 +1118,17 @@ components:
             - inactive
             - development
             - future
+    features:
```
davidgamez, line 1121: [suggestion]: Even though this is an array, I would name it `feature` for consistency, as we already have the filter `status`. It would be better to have them all in plural, but the decision was made already for the first parameters, and we shouldn't break backward compatibility.

web-app/src/app/screens/Feeds/index.tsx
qcdyx: Done.

api/src/feeds/impl/models/search_feed_item_result_impl.py
qcdyx: good catch, removed.

davidgamez:
Out of scope of this: #1156

web-app/src/app/screens/Feeds/index.tsx
```diff
@@ -125,6 +125,7 @@ export default function Feed(): React.ReactElement {
             // Fixed status values for now, until a status filter is implemented
             // Filtering out deprecated feeds
             status: ['active', 'inactive', 'development', 'future'],
+            feature: selectedFeatures,
```
Alessandro100, line 128: we have a variable `areFeatureFiltersEnabled` if they aren't we shouldn't set the selectedFeatures

```
feature: areFeatureFiltersEnabled ? selectedFeatures : undefined
```

Alessandro100: This isn't in the scope of this ticket, but we should also do this for `isOfficialTagFilterEnabled`

davidgamez: > This isn't in the scope of this ticket, but we should also do this for `isOfficialTagFilterEnabled`

If there is a risk of releasing without the condition, it should be part of the scope of the issue.

Alessandro100: I could make another ticket to address it but since it's a 1 liner it would be easier to address here
```
is_official: isOfficialTagFilterEnabled ? (isOfficialFeedSearch || undefined) : undefined
```

davidgamez: I would prefer it here, as leaving it as it is, will block a potential release.

davidgamez (approved):
LGTM!

Issue #1129, opened 2025-04-22
Add versions OR filter


emmambd:
### Describe the problem

 add versions filter to search endpoint 

### Proposed solution

See above

PR #1162, opened 2025-04-30
feat: added versions filtering to search

cka-y:
**Summary:**
- Added `versions` field to the `FeedSearch` materialized view and exposed it in the search API response for GBFS feeds
- Introduced a new optional `versions` query parameter to the `/search` endpoint  
  → Accepts a comma-separated list (e.g. `?versions=2.2,3.0`)  
  → Filters feeds that support **any** of the specified versions (logical OR)

**Notes:**
- Version filtering is only applied when the `versions` parameter is provided and non-empty
- Supports exact matches on version strings as returned by the backend (e.g. `"2.2"`, `"3.0"`)

**Testing tips:**
Run API locally.
```http
GET /v1/search?data_type=gbfs&versions=2.2,3.0
``` 

### DIFF
--- a/api/src/feeds/impl/search_api_impl.py
+++ b/api/src/feeds/impl/search_api_impl.py
@@ -2,7 +2,7 @@ from typing import List
 
 from sqlalchemy import func, select
 from sqlalchemy.orm import Query, Session
-
+from sqlalchemy.dialects.postgresql import array
 from shared.database.database import Database, with_db_session
 from shared.database.sql_functions.unaccent import unaccent
 from shared.database_gen.sqlacodegen_models import t_feedsearch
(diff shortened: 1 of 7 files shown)


emmambd:
@cka-y Looks like this solves #1129 too?

docs/DatabaseCatalogAPI.yaml
```diff
@@ -1335,6 +1342,15 @@ components:
       schema:
         type: string
 
+    versions_query_param:
+      name: versions
```
davidgamez, line 1346: [suggestion]: Even though this is an array, I would name it version for consistency, as we already have the filter status and feature. It would be better to have them all in plural, but the decision was made already for the first parameters, and we shouldn't break backward compatibility.

api/src/feeds/impl/search_api_impl.py
```diff
@@ -122,12 +130,13 @@ def search_feeds(
         feed_id: str,
         data_type: str,
         is_official: bool,
+        versions: str,
         search_query: str,
-        features: List[str],
+        feature: List[str],
```
davidgamez, line 135: 👍

docs/DatabaseCatalogAPI.yaml
cka-y: done ✅

davidgamez (approved):

PR #1163, opened 2025-04-30
feat: Add gbfs endpoint to the GBFSVersion entity

### LINKED ISSUE #1124: Add `gbfs` endpoint to the `GBFSVersion` entity
### Describe the problem

In `gbfs.json`, there are two cases:

1. The `gbfs` endpoint is explicitly listed — we register it ✅  
2. The `gbfs` endpoint isn’t listed — but we still process all the downstream endpoints just fine.  
   The problem is: we **don’t register** the `gbfs` endpoint itself because it isn't explicitly listed, even though we *should*.

### Proposed solution

We already have the `url` stored in the `GBFSVersion` entity. Let’s fix the gap:

- If the `gbfs` endpoint is missing from `gbfs.json`, create a record for it under the existing `url` (as long as it’s not already registered) [This can be done with SQL no need for additional code].
- Update the `gbfs_validator` cloud function so that it **automatically registers** the `gbfs` endpoint when it’s missing, even if it’s not listed explicitly.

Other consideration: Add a `url`/`autodiscovery_url` field to the `versions` field in the API response.

davidgamez:
**Summary:**
Closes #1124 
This PR adds the autodiscovery URL to the list of endpoints when it's not explicitly defined in the GBFS response per version. Also, populate the current data in PROD. 

**Expected behavior:** 

The autodiscovery URL is always persisted in the DB as a GBFS endpoint.

**Testing tips:**
SQL update executed in DEV and executed the gbfs-batch function in DEV, the following query verified that there are no gbfs versions without the `gbfs` endpoint:
```
SELECT feed.stable_id, gbfsversion.*
FROM gbfsversion
inner join feed on gbfsversion.feed_id = feed.id
WHERE NOT EXISTS (
    SELECT 1
    FROM gbfsendpoint ge
    WHERE ge.gbfs_version_id = gbfsversion.id
      AND ge.name = 'gbfs'
);
```
GBFS endpoint added to a feed that don't have GBFS endpoint listed in version 1.1:

Screenshot 2025-05-01 at 4 05 02 PM

### DIFF
--- a/functions-python/gbfs_validator/src/gbfs_data_processor.py
+++ b/functions-python/gbfs_validator/src/gbfs_data_processor.py
@@ -114,6 +114,13 @@ class GBFSDataProcessor:
                 language = None
             print(language)
             endpoints += GBFSEndpoint.from_dict(feed_match.value, language)
+
+        # If the autodiscovery endpoint is not listed then add it
+        if not any(endpoint.name == "gbfs" for endpoint in endpoints):
+            endpoints += GBFSEndpoint.from_dict(
+                [{"name": "gbfs", "url": gbfs_json_url}], None
+            )
+
         unique_endpoints = list(
             {
                 f"{endpoint.name}, {endpoint.language or ''}": endpoint
--- a/functions-python/gbfs_validator/src/gbfs_data_processor.py
+++ b/functions-python/gbfs_validator/src/gbfs_data_processor.py
@@ -284,9 +284,9 @@ class GBFSDataProcessor:
         features: List[str],
     ) -> Gbfsendpoint:
         """Update or create a GBFS endpoint entity."""
-        formatted_id = (
-            f"{self.stable_id}_{version}_{endpoint.name}_{endpoint.language or ''}"
-        )
+        formatted_id = f"{self.stable_id}_{version}_{endpoint.name}"
+        if endpoint.language:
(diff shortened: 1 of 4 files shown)


functions-python/gbfs_validator/src/gbfs_data_processor.py
```diff
@@ -277,9 +284,9 @@ def update_or_create_gbfs_endpoint(
         features: List[str],
     ) -> Gbfsendpoint:
         """Update or create a GBFS endpoint entity."""
-        formatted_id = (
-            f"{self.stable_id}_{version}_{endpoint.name}_{endpoint.language or ''}"
-        )
+        formatted_id = f"{self.stable_id}_{version}_{endpoint.name}"
```
jcpitre, line 287: Can you explain why in your screenshot the last item (`zaragoza_1.1_gbfs_`) ends with an underscore?
Whereas here it should not finish with an underscore.

davidgamez: The screenshot was taken before [commit](https://github.com/MobilityData/mobility-feed-api/pull/1163/commits/3ab8656b9100000ceb4751d9ac93511884f67eb0)

jcpitre (approved):
LGTM!

PR #1265, opened 2025-06-17
fix: 2 corrections on gbfs populate

### LINKED ISSUE #1242: GBFS feeds are not being removed from API
### Describe the bug

For example: Beryl London was removed from systems.csv in January 2025 due to a permanent shutdown ([PR](https://github.com/MobilityData/gbfs/pull/717)), but it is still present in the Mobility Database ([feed page](https://mobilitydatabase.org/feeds/gbfs/gbfs-beryl_london)).
It may be confusing for users that the number of GBFS feeds in the [Mobility Database](https://mobilitydatabase.org/feeds?gbfs=true) (1280 feeds) is different from the number in [systems.csv](https://github.com/MobilityData/gbfs/blob/master/systems.csv) (1227 feeds).

Other feed example that should be deleted: https://mobilitydatabase.org/feeds/gbfs/gbfs-targumures

### Steps/Code to Reproduce

See links above

### Expected Results

When a feed is not in systems.csv it does not appear in the Mobility Database

### Actual Results

Feed appears

### Screenshots

_No response_

### Files used

_No response_

### Additional notes

_No response_

### LINKED ISSUE #1268: Modify DB schema to handle entity deletion properly
In #1265, we marked a GBFS feed as deprecated if it was removed from `systems.csv`.

However, our actual intent is to **fully remove the feed from the database** in that case.

Just deleting the feed with `session.delete(gbfs_feed)` does not work because of constraint that prevent deletion of a Feed if it's present in the externalid table.

Intially we tried:
```python
      gbfs_feed.externalids.clear()
      session.delete(gbfs_feed)
```
This worked temporarily, but only because our test feeds didn’t have all the relationships (e.g., `entitytypefeed`) that production feeds do. Moreover, this approach is fragile because it relies on the Python code being aware of all database constraints and relationships.

We also tried modifying [`database.py`](https://github.com/MobilityData/mobility-feed-api/blob/1ee3c9a9b56d78f933bb1735a03830234aed6e96/api/src/shared/database/database.py#L64) to augment `cascade_entities`, but it didn’t have the desired effect.

but it worked only because the test feeds we were using did not have all the relationships a production feed has (e.g. the entitytypefeed). Also this code is not very good because it assumes the python code is aware of all the constraints.

We tried modifying [database.py](https://github.com/MobilityData/mobility-feed-api/blob/1ee3c9a9b56d78f933bb1735a03830234aed6e96/api/src/shared/database/database.py#L64) to augment `cascade_entities`, but it did not work.

Ultimately, we got the deletion working using the following Liquibase change:
```
ALTER TABLE ExternalID DROP CONSTRAINT IF EXISTS externalid_feed_id_fkey;

ALTER TABLE ExternalID
ADD CONSTRAINT externalid_feed_id_fkey
FOREIGN KEY (feed_id) REFERENCES Feed(id) ON DELETE CASCADE;
```
This solves the problem only for the `feed.externalids` relationship. We now need to review the database schema and apply similar `ON DELETE CASCADE` logic to all relationship tables where the data doesn't make sense without the referenced Feed.

jcpitre:
Closes #1242 
Closes #1268 

The two corrections are:

- Deleted feeds from the DB if removed form systems.csv
   - There was code that would filter out feeds that were in in the DB but not in systems.csv. Thus we did not know that some feeds were removed from systems.csv
-  Stopped processing feeds based on changes in location
   - The location is initially set with the populate scripts from systems.csv, but then it is changed by the function that determines the location from the data. So comparing locations to see if something changed in CSV did not work.

For deletion, this PR also setup cascade deletes on certain DB tables in the SQL schema. For example, id a GbfsFeed is deleted, the entry in gbfsversion table referring to this feed is also deleted. This prevent failing of the deletion because gbfsversion.feed_id cannot be null and is a foreign key to gbfsfeed.

**Testing tips:**
See test_populate_gbfs.py for the tests that have been done.
We miodified the way the DB is filled in the api integration tests to allow 2 .csv files to be loaded. This allowed to test loading a csv in a DB that is already populated, as was necessary for this PR.

### DIFF
--- a/api/src/scripts/gbfs_utils/comparison.py
+++ b/api/src/scripts/gbfs_utils/comparison.py
@@ -47,6 +47,7 @@ def compare_db_to_csv(df_from_db, df_from_csv, logger):
         return None, None
 
     # Align both DataFrames by "System ID"
+    # Keep the System ID column because it's used later in the code
     df_from_db.set_index("System ID", inplace=True, drop=False)
     df_from_csv.set_index("System ID", inplace=True, drop=False)
 
@@ -85,6 +86,7 @@ def compare_db_to_csv(df_from_db, df_from_csv, logger):
             logger.info(80 * "-")
 
     # Merge differing rows with missing_in_db to capture all new or updated feeds
+    # Drop the index because we have it as the System ID column.
     all_differing_or_new_rows = pd.concat([differing_rows, missing_in_db]).reset_index(drop=True)
 
     return all_differing_or_new_rows, missing_in_csv
--- a/api/src/scripts/gbfs_utils/comparison.py
+++ b/api/src/scripts/gbfs_utils/comparison.py
@@ -5,9 +5,7 @@ from shared.database_gen.sqlacodegen_models import Gbfsfeed
 
 def generate_system_csv_from_db(df, db_session):
     """Generate a DataFrame from the database with the same columns as the CSV file."""
-    stable_ids = "gbfs-" + df["System ID"]
     query = db_session.query(Gbfsfeed)
-    query = query.filter(Gbfsfeed.stable_id.in_(stable_ids.to_list()))
     query = query.options(
         joinedload(Gbfsfeed.locations), joinedload(Gbfsfeed.gbfsversions), joinedload(Gbfsfeed.externalids)
     )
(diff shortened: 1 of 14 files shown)


api/src/scripts/gbfs_utils/comparison.py
```diff
@@ -68,7 +67,11 @@ def compare_db_to_csv(df_from_db, df_from_csv, logger):
     common_ids = df_from_db.index.intersection(df_from_csv.index)
     df_db_common = df_from_db.loc[common_ids]
     df_csv_common = df_from_csv.loc[common_ids]
-    differences = df_db_common != df_csv_common
+
+    # Exclude 'Location' from comparison because the DB values might have been changed in the
+    # python function that calculates the location.
+    columns_to_compare = [col for col in df_db_common.columns if col != "Location"]
```
cka-y, line 73: 👍

cka-y (approved):
LGTM

PR #1529, opened 2025-12-18
chore: add liquibase changelog composite primary key

davidgamez:
**Summary:**

Previously, DEV had an incorrect single-column primary key on filename, which was too restrictive and caused issues when multiple XML changelogs reused the same filename. QA and PROD, on the other hand, had no primary key at all, which meant the same changeset could theoretically be applied more than once.

By removing the single-column PK and adding the proper composite PK, all environments now behave consistently and safely:
- Prevents duplicate execution of the same changeset
- Supports multiple XML changelog files with identical filenames
- Aligns all environments with Liquibase’s expected schema

A possible root cause is that Liquibase doesn't update its own tables, and DEV, QA, and PROD environments might have been created with different Liquibase versions.
Here is an interesting thread on why recent Liquibase versions no longer enforce a primary key, [link](https://forum.liquibase.org/t/why-does-databasechangelog-not-have-a-primary-key/3270/6)

**Expected behavior:** 

Liquibase should complete successfully in all environments.

**Testing tips:**

Liquibase should complete successfully in all environments.

### DIFF
--- a/api/src/scripts/populate_db.py
+++ b/api/src/scripts/populate_db.py
@@ -45,7 +45,7 @@ class DatabasePopulateHelper:
     Helper class to populate the database
     """
 
-    def __init__(self, filepaths):
+    def __init__(self, filepaths, test_mode=False):
         """
         Specify a list of files to load the csv data from.
         Can also be a single string with a file name.
@@ -60,7 +60,10 @@ class DatabasePopulateHelper:
             filepaths = [filepaths]
 
         for filepath in filepaths:
-            new_df = pandas.read_csv(filepath, low_memory=False)
+            if test_mode:
+                new_df = pandas.read_csv(filepath, comment="#", low_memory=False)
+            else:
+                new_df = pandas.read_csv(filepath, low_memory=False)
             self.df = pandas.concat([self.df, new_df])
 
         self.filter_data()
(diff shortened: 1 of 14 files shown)


jcpitre:
Just an idea:
One thing that I find annoying a bit in our current setup is that files specified in changelog.xml are not in alphabetical order, as is given when doing a simple ls. This makes it a bit harder to find the order of application of each file since we need to refer to the changelog.xml. 
Since we're playing with databasechangelog, would it be a good time now to rename files so they are in alphabetical order in changelog? We could for example put a prefix like 001, 002 on each file.

davidgamez:
> Just an idea: One thing that I find annoying a bit in our current setup is that files specified in changelog.xml are not in alphabetical order, as is given when doing a simple ls. This makes it a bit harder to find the order of application of each file since we need to refer to the changelog.xml. Since we're playing with databasechangelog, would it be a good time now to rename files so they are in alphabetical order in changelog? We could for example put a prefix like 001, 002 on each file.

The schema update pipeline is working well and doesn’t have issues today. Renaming the files would change the changelogs and their IDs, which could cause all updates to run again once the change is merged. The only real benefit would be a cleaner folder order, which isn’t something humans usually need to read.

Also, we often have several PRs updating the schema at the same time. It’s not clear how we would choose filenames in a way that avoids conflicts. For these reasons, I’m not in favor of renaming the files in the changelog folder.

jcpitre (approved):

PR #1737, opened 2026-06-11
feat: Gtfs change tracker 1634

jcpitre:
closes #1634 

This pull request introduces a new GTFS change tracking feature to the pipeline. It adds a new Cloud Function (`gtfs-change-tracker`) that computes and stores structured diffs between pairs of GTFS datasets, integrates this function into the batch processing workflow, and updates the infrastructure to support the new service. The changes span Python code, requirements, and Terraform infrastructure.

**GTFS Change Tracking Feature:**

* Added a new Cloud Function `gtfs-change-tracker` that computes diffs between two GTFS datasets, uploads the changelog to GCS, and persists a summary in the database. This includes the main logic (`src/main.py`), configuration (`function_config.json`), and dependencies (`requirements.txt`, `requirements_dev.txt`). [[1]](diffhunk://#diff-59933ca41c9cef5109b3dc353a8ed3a82fd3e218e59d1d7e638017629d016f8eR1-R326) [[2]](diffhunk://#diff-14482f7f1de183c0abb4aad9a1ada2beb41289d31913f71254f9a81c0bf21696R1-R31) [[3]](diffhunk://#diff-3682b7b08d67d4318267df55b17165ef2a995d24b576d696ea31963bf4651636R1-R22) [[4]](diffhunk://#diff-4cdbb00ff04a52c5fd2af448ebe878d27c27dd24ee7b60f99bbbd8387b53b308R1-R2)

* Implemented a utility function `create_http_gtfs_change_tracker_task` to enqueue Cloud Tasks for the new function, and integrated it into the pipeline so that a change tracking task is created when a new dataset is processed and a previous dataset exists. [[1]](diffhunk://#diff-472bd681a31f89d978dff0c7d5c3e4490bc18ef226b06bf45c2c8f7c5867a234R538-R571) [[2]](diffhunk://#diff-03a569e1801d8f2e5d97a4b049895619c997aee23874645bb401a8f197db91f3L11-R15) [[3]](diffhunk://#diff-03a569e1801d8f2e5d97a4b049895619c997aee23874645bb401a8f197db91f3R143-R164)

**Infrastructure and Deployment:**

[...]

PR #1743, opened 2026-06-19
fix: duplicate some Cloud Run IAM grants to functions-python module

jcpitre:
### Problem
This was a new error that happened after merging #1737 

 The `datasets-batch-deployer-dev` workflow was failing with:

Error 404: Resource 'gtfs-datasets-comparer-dev' of kind 'SERVICE' ... does not exist.

 `infra/batch/main.tf` had two `google_cloud_run_service_iam_member` resources that granted `roles/run.invoker` on the `pmtiles-builder-dev` and `gtfs-datasets-comparer-dev` Cloud Run services by **hardcoded name**. These services are deployed by a separate workflow (`api-dev`/`infra/functions-python`), so if they don't exist yet — as was the case after renaming `gtfs-change-tracker` → `gtfs-datasets-comparer` — Terraform fails trying to set IAM on a non-existent resource.
 
 ### Solution
 Moved the two `google_cloud_run_service_iam_member` resources from `infra/batch/main.tf` into `infra/functions-python/main.tf`, co-located with the `google_cloudfunctions2_function` resources that actually create those services. Using `google_cloudfunctions2_function.X.name` as the `service` reference creates an implicit dependency, ensuring the Cloud Run service exists before IAM is applied. This follows the same pattern already used for the `tokens` and `operations_api` functions.
 
 Also removed the now-unused `function_pmtiles_builder_config` and `function_gtfs_datasets_comparer_config` locals from `infra/batch/main.tf`.
 
### DIFF
--- a/infra/batch/main.tf
+++ b/infra/batch/main.tf
@@ -42,6 +42,10 @@ locals {
   retention_duration_seconds = lower(var.environment) == "prod" ? 2678400 : 604800
 
   deployment_timestamp = formatdate("YYYYMMDDhhmmss", timestamp())
+
+  function_pmtiles_builder_config = jsondecode(file("${path.module}/../../functions-python/pmtiles_builder/function_config.json"))
+
+  function_gtfs_datasets_comparer_config = jsondecode(file("${path.module}/../../functions-python/gtfs_datasets_comparer/function_config.json"))
 }
 
 data "google_vpc_access_connector" "vpc_connector" {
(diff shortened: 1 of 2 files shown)


infra/batch/main.tf
```diff
@@ -485,18 +481,3 @@ resource "google_compute_global_forwarding_rule" "files_http_lb_rule_ipv4" {
...
 
-resource "google_cloud_run_service_iam_member" "pmtiles_builder_invoker" {
-  project  = var.project_id
-  location = var.gcp_region
-  service  = "${local.function_pmtiles_builder_config.name}-${var.environment}"
-  role     = "roles/run.invoker"
-  member   = "serviceAccount:${google_service_account.functions_service_account.email}"
-}
-
-resource "google_cloud_run_service_iam_member" "gtfs_datasets_comparer_invoker" {
```
davidgamez, line 496: Is this not affecting the pmtiles builder queue?

davidgamez:
I think we need the iam permission in both TF files rather than moving the role from one to another.

jcpitre:
> I think we need the iam permission in both TF files rather than moving the role from one to another.

We could have them in both. But there would still be a problem since pmtiles_builder_invoker would still refer to a service that is created in python-function. See [here](https://github.com/MobilityData/mobility-feed-api/blob/bed9b9c7120bb5f127f22d284206a716f751c01b/infra/batch/main.tf#L491)

davidgamez:
> > I think we need the iam permission in both TF files rather than moving the role from one to another.
> 
> We could have them in both. But there would still be a problem since pmtiles_builder_invoker would still refer to a service that is created in python-function. See [here](https://github.com/MobilityData/mobility-feed-api/blob/bed9b9c7120bb5f127f22d284206a716f751c01b/infra/batch/main.tf#L491)

The issue is that if the permission is not present, the Cloud Task will fail. We had planned(not done yet...) to merge these two TF files so we can avoid this cross-reference issue (completely out of scope in this PR)

jcpitre:
> > > I think we need the iam permission in both TF files rather than moving the role from one to another.
> > 
> > 
> > We could have them in both. But there would still be a problem since pmtiles_builder_invoker would still refer to a service that is created in python-function. See [here](https://github.com/MobilityData/mobility-feed-api/blob/bed9b9c7120bb5f127f22d284206a716f751c01b/infra/batch/main.tf#L491)
> 
> The issue is that if the permission is not present, the Cloud Task will fail. We had planned(not done yet...) to merge these two TF files so we can avoid this cross-reference issue (completely out of scope in this PR)

I put back the permissions in batch

davidgamez (approved):

PR #1773, opened 2026-07-23
feat: add date-window and validator-version filters to rebuild_missing_validation_reports 

### LINKED ISSUE #1762: Re-run Validator 8.0.1 on datasets from March - May
### Describe the problem

Due to [the service date window bug](https://github.com/MobilityData/gtfs-validator/issues/2017) that was introduced in the validator, our Seal of Reliability calculation will be wrong for historical datasets that were run before 8.0 was introduced.

The calculation is based on the last six months worth of datasets for 1 feed.

### Proposed solution

We need to re-run the validator on any datasets published between March and May of 2026 so our Seal of Reliability calculation will be accurate by the time of the announcement in end of September. 

_No response_

### Additional context

_No response_

cka-y:
**Summary:**

Extends `rebuild_missing_validation_reports` (`functions-python/tasks_executor`) by adding optional filters so we can re-validate a historical window against the prod validator. Closes #1762.

- `filter_downloaded_after` / `filter_downloaded_before`: ISO-date window on `downloaded_at`.
- `filter_validator_version_prefix`: target datasets lacking **any** report matching the prefix (e.g. `"8."`), not just a non-exact version.
- `include_all_datasets`: consider **all** datasets of a feed, not just the latest.

All default to preserve existing behavior. README + unit tests updated.

**Expected behavior:**

Dry run reports the count of published feeds' datasets in the window missing an `8.*` report; non-dry run triggers a validation workflow per candidate against prod and writes the reports. Response `params` echo the applied filters.

**Testing tips:**

Fields must be nested under `"payload"`. Dry run first:

```json
{
  "task": "rebuild_missing_validation_reports",
  "payload": {
    "dry_run": true,
    "filter_downloaded_after": "2026-03-01",
    "filter_downloaded_before": "2026-06-01",
    "filter_validator_version_prefix": "8.",
    "include_all_datasets": true,
    "validator_endpoint": "https://gtfs-validator-web-mbzoxaljzq-ue.a.run.app"
  }
}
```

Check `params` echo your filters and review `total_candidates`, then flip `dry_run` to `false` to trigger.

- `include_all_datasets: true` is what covers all datasets in the window (else latest-only).
- Count filters by the feed's `operational_status` (default published) with an exclusive upper bound, so it won't match a raw `downloaded_at BETWEEN` query.
- Idempotent: already-triggered and already-`8.*` datasets are skipped. Try `"limit": 10` for a small batch first.
Notes for testers:
- The count reflects each feed's **latest** dataset only, `operational_status = 'published'`, and an **exclusive** upper bound — so it won't match a raw `SELECT ... FROM gtfsdataset WHERE downloaded_at BETWEEN ...` count.
- Re-running is idempotent: already-triggered datasets are skipped, and any dataset that already has an `8.*` report is excluded.
- Try omitting the new fields to confirm the original behavior is unchanged, and try `"limit": 10` for a small end-to-end batch first.

### DIFF
--- a/functions-python/tasks_executor/src/tasks/validation_reports/README.md
+++ b/functions-python/tasks_executor/src/tasks/validation_reports/README.md
@@ -63,6 +63,12 @@ sequenceDiagram
 Finds GTFS datasets that are missing a validation report **or** have a report from an
 older validator version, then triggers a GCP Workflow for each one.
 
+By default "older version" means "not the exact version returned by the validator
+endpoint". Set `filter_validator_version_prefix` (e.g. `"8."`) to instead target
+datasets that lack **any** report matching that major-version prefix — useful when you
+want to re-validate everything missing an `8.*` report regardless of the exact patch
+version the endpoint reports.
+
 The task is **resumable**: if it times out mid-loop, calling it again skips datasets
 that were already triggered (tracked in `task_execution_log`).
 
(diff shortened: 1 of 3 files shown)


davidgamez (approved):
LGTM!

PR #1790, opened 2026-07-30
fix: operations api async calls

davidgamez:
## Summary

The Operations API's generated FastAPI routers invoke each impl method **without `await`**
(`return BaseXApi.subclasses[0]().method(...)`). Several impl methods were declared `async def`,
so at runtime they returned un-awaited coroutines instead of responses — producing HTTP 500
(`ResponseValidationError`) on the live API. Unit tests missed it because they call the impl
directly and `await` it, never crossing the router→impl seam.

This PR:

- **Fixes the 500s** by making the affected impl methods (and the `validate_request` wrapper)
  plain `def`, so the routers' non-awaited calls return real responses. This is a correctness fix,
  **not** a performance cleanup — and it does **not** reduce concurrency. Requests are still served
  concurrently by the Cloud Run / functions-framework worker threads; only the Python function
  *definitions* change from `async def` to `def`.
- **Removes phantom update changes** in `detect_changes`: excludes derived/read-only fields
  (`source_info.license_is_spdx`) and normalizes "absent" values (`None` / `""` / `[]`) on both
  sides of the diff, so an unchanged feed saved from Retool no longer reports false edits.
- **Adds regression guards**: a static `test_routes_are_sync.py` asserting no route-backing method
  is a coroutine, and an `operations-api-surface` CI job that boots the real API and hits every
  endpoint over HTTP to catch any 5xx from a coroutine/serialization leak.

> Note on terminology: "sync" / "removing async" here refers to the function *definitions* only.
> The endpoints are not becoming single-threaded and do not lose concurrency.

**Expected behavior:** 

**Testing tips:**


cka-y (approved):
LGTM

Issue #1784, opened 2026-07-29
Seal of reliability: Implement other criteria evaluation


jcpitre:
From #1761 
#1783 will implement the `official` criterion
#1782 will implement the `fresh-continuous-coverage` criterion

[Logic spike calculation breakdown
](https://github.com/MobilityData/mobility-feed-api/issues/1761#issuecomment-4987530241)
We need to implement the other criteria:

- Stable
- Available
- Compliant
- Fresh-future-coverage

We bundle them together here because they should be simple to implement.

Issue #1782, opened 2026-07-28
Seal of Reliability: Implement the Fresh-continuous-coverage criterion


jcpitre:
This is a spin-off from #1761 that gives details on the Fresh-Continuous-Coverage criterion.
Here is the proposed algorithm:
* If feed date not available  
  * If calendar range \< 2 years  
    * Calendar gap (1 day) \- F  
      * Calendar no gap \- T  
    * If calendar range \> 2 years \- F  
* If feed date available  
    * If feed date range \< 2 years  
      * If feed date no gap  
        * Calendar gap \- T  
      * If feed date gap (any number of days)   
        * Calendar gap \- F  
        * Calendar no gap T  
    * If feed date range \> 2 years \- F

- Rationale behind 2 years   
  - 80% of feeds with feed info file had a date range that covered 2 years or less 

This algorithm is adequate since there is no grace period for this criterion, meaning there is no way to correct the error.
If there was we would need to make sure there is no gap between the latest dataset and any previous dataset to leave a chance to correct the error within the grace period.

Issue #1761, opened 2026-07-15
[Spike] Logic to update the Seal of Reliability


Alessandro100:
Write a shared function that would update the feed.seal_of_reliability property. With tests

Current Logic (As of 2026-07-27)
> Official:
> - Official at time of reviewing dataset. Does not require a 6 month check.
> - Calculation: is_official = true
>
> Stable:
> - dataset created_at has to be six months or longer in the past AND it is_producer_unstable = null/false.
>
> Available:
> - Successfully returned a response for 6 months. No 15 subsequent days within this period of outages.
> - Calculation: Every day, look back 6 months (or as long as we have availability data) and check that there are no 15 subsequent days of outage. Success = returned a response in the HTTP 200s. [From pre-existing availability table]
> - Grace period: 14 days. So if a feed fails to return a successful response, they have 14 days to fix the problem before disqualification.
>
> Compliant:
> - Returns no validation errors for 6 months.
> - Calculation: 0 errors returned from the validation report. Any validation error that is currently in the feed is less than 30 days old.
> - Grace period: If a validation error is returned, there are 30 days for it to be fixed before it's disqualified.
>
> Fresh: Includes 7 days or more service coverage each day.
> - Calculation: service_end_date extends 7 day or more into the future.
> - Grace period: if service_end_date extends less than 7 days into the future, 14 days to fix it before disqualification.
> - Feeds marked as seasonal should not run this calculation.
>
> Fresh: Has continuous service coverage across subsequent datasets with no coverage gaps.
> Calculation: [View in-depth calculation here.](https://docs.google.com/document/d/16_B5Bfw1LIQ2VbitlUny0PBERcMOTvXa_2D0LHS30D0/edit?tab=t.qm11f6a1hs8s)

For the Fresh-Continuous Coverage, here is the algorithm:
* If feed date not available  
  * If calendar range \< 2 years  
    * Calendar gap (1 day) \- F  
      * Calendar no gap \- T  
    * If calendar range \> 2 years \- F  
  * If feed date available  
    * If feed date range \< 2 years  
      * If feed date no gap  
        * Calendar gap \- T  
      * If feed date gap (any number of days)   
        * Calendar gap \- F  
        * Calendar no gap T  
    * If feed date range \> 2 years \- F

- Rationale behind 2 years   
  - 80% of feeds with feed info file had a date range that covered 2 years or less

emmambd:
[...]

Issue #1760, opened 2026-07-15
Add Seal of Reliability properties to schema


Alessandro100:
Add the decided properties to best represent the seal of reliability to the schema 

!important: we must really get on top of naming because once it's in the schema its going to be harder to change

jcpitre:
The suggestion is to create two new tables that aggregate all information relating to the seal of reliability.
 See #1761 for the usage of these tables.

As of 2026-08-19, there was some changes in the tables. The new tables can be found there:  [feat_1783.sql](https://github.com/MobilityData/mobility-feed-api/blob/main/liquibase/changes/feat_1783.sql)

 ### New Tables - Replaced in [feat_1783.sql](https://github.com/MobilityData/mobility-feed-api/blob/main/liquibase/changes/feat_1783.sql)
 
 The `feed_reliability_seal` and `seal_criterion` tables intentionally denormalize data already present in `gtfs_feed_availability_check`, `validationreport`, and `gtfsdataset`. This follows the same pattern already used in this codebase (e.g. `gtfsfeed.latest_dataset_id`): raw tables are the source of truth and append-only; the seal tables are a nightly-computed cache that lets the API answer "does this feed have the seal?" in a single row lookup with no joins.
 
 `feed_reliability_seal` holds one row per feed and owns the overall seal outcome. `seal_criterion` holds one row per feed per criterion (six rows per feed) and owns all per-criterion state. Out of scope for this issue, but this would make adding (or removing) some criteria in the future easier and maybe qualify a given criterion, e.g. having an experimental flag. 
 
  ```sql
 CREATE TABLE feed_reliability_seal (
     feed_id        VARCHAR(255) PRIMARY KEY
                        REFERENCES gtfsfeed(id) ON DELETE CASCADE,
 
     has_seal       BOOLEAN NOT NULL DEFAULT FALSE,
     seal_earned_at TIMESTAMPTZ,
     seal_lost_at   TIMESTAMPTZ,
 
     created_at     TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     updated_at     TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
 
 CREATE TYPE seal_criterion_name AS ENUM (
     'official',
     'stable',
     'available',
     'compliant',
     'fresh_coverage',
     'fresh_continuous'
 );
 
 -- raw_* columns reflect the instantaneous state at the last evaluation, with no grace applied.
[...]

PR #1795, opened 2026-08-05
feat: Seal of reliability infrastructure and official criterion

### LINKED ISSUE #1783: Seal of reliability: Implement logic and Official criterion
Implement the logic described in https://github.com/MobilityData/mobility-feed-api/issues/1761#issuecomment-4987530241 and use the `official` criterion for the initial development.

jcpitre:
Closes #1783 

Implements the seal or reliability algorithm as defined in #1761, but uses only `official` as the first criterion implemented.

The code is placed in `functions-python/task_executor` as the `seal_of_reliability` task.

You need to specify the list of feeds to process, listed in the `stable_feed_ids` parameter.
To which this filter is applied (as defined in `functions-python/tasks_executor/src/tasks/seal_of_reliability/context.py`):

```
db_session.query(Gtfsfeed).filter(
        Feed.data_type == "gtfs",
        Feed.status.notin_(["deprecated", "development"]),
        Feed.operational_status == "published",
    )
``` 
See README.md in the tasks_executor folder for the list of parameters.

# From our AI friend

## What this does

Implements the Seal of Reliability evaluation logic and the **Official** criterion, per #1783. The logic is described in #1761; the `sealcriterion` and `feedreliabilityseal` tables come from #1760 (deployed via #1792).

A new `tasks_executor` task, `update_seal_of_reliability`, evaluates a given list of GTFS feeds and writes the two seal tables. It always runs against an explicit `stable_feed_ids` list — there is no run-the-whole-catalogue mode; eligibility (gtfs, published, not deprecated/development) is applied on top of the list. The source tables (`feed`, `gtfsdataset`, `validationreport`, `gtfs_feed_availability_check`) are only ever read.

The remaining five criteria are #1784 and #1782. Each is a new evaluator subclass plus whatever data it needs on the context.

## Schema change

`feat_1783.sql` alters the #1760 tables. They are deployed but empty — the only writer is the task in this PR — so the columns are replaced in place, with no data to migrate.

- The two status columns become a `seal_criterion_status` enum: `pass | fail | unknown | never_evaluated | not_applicable`. As booleans they could only say pass/fail/NULL, and NULL had to mean three different things (never evaluated, not evaluable this run, not applicable) that the roll-up now handles in different directions.
- `last_verdict_at` is added — the last `pass`/`fail`. `evaluated_at` now means the last *attempt*, whatever it returned. The gap between them is how long a criterion has been stuck without a verdict.
- The three failure timestamps are renamed (`*_raw_* → *_observed_*`, `*_grace_* → *_confirmed_*`); `probation_start` is added.

## How to run it

```json
{ "task": "update_seal_of_reliability", "payload": { "dry_run": true, "stable_feed_ids": ["mdb-1210"] } }
```

`stable_feed_ids` is **required** and non-empty. `dry_run` defaults to **true**. Other parameters: `limit`, `criteria`, `batch_size`, `now`, `max_reported_feeds` — documented in `functions-python/tasks_executor/README.md` with a local curl walkthrough.

No Cloud Scheduler job yet; the task is invoked manually. When scheduled it must run after `check_gtfs_feed_availability`, since the Available criterion (#1784) reads that day's availability rows.

## Structure

| File | Role |
|---|---|
| `criteria.py` | `SealCriterionName`, the `CriterionStatus` and `CriterionPhase` enums, `PROBATION_PERIOD` |
| `context.py` | Feed eligibility query and the per-feed `FeedSealContext`, built in bulk per batch |
| `evaluators/base.py` | `CriterionEvaluator` base class, `CriterionObservation` |
| `evaluators/official.py` | The Official criterion |
| `state_machine.py` | `transition()` — steps 1-4; `phase()` — the derived phase |
| `seal_updater.py` | `update_seals()` — orchestration, the `has_seal` roll-up, persistence |
| `update_seal_of_reliability.py` | Task entry point: payload parsing only |

Evaluators are pure functions over a pre-loaded context and never query, which keeps the query count proportional to the number of criteria, not feeds.

## Decisions worth a reviewer's attention

**`unknown` and `not_applicable` go opposite ways.** `unknown` (inputs missing) keeps the stored `confirmed_status`, so an upstream outage freezes a criterion rather than waiving it. `not_applicable` (excluded for this feed, e.g. a seasonal feed for a coverage criterion) overwrites `confirmed_status` and drops the criterion from the roll-up. **This is a behaviour change worth confirming against product intent**: an excluded criterion can now hand a feed its seal back.

**Probation suspends the grace period.** A criterion serving probation gets no grace on its next failure — it is confirmed the same day. No seal outcome changes (a criterion on probation already denies the seal), but the recorded status is now honest on the day the check failed. This also makes `in_grace_period` and `on_probation` mutually exclusive, so `phase` needs no precedence rule.

**`phase` is derived, not stored.** `steady | in_grace_period | on_probation` is a pure function of `probation_start`, `confirmed_status` and `first_observed_failure_at` — computed on read, no column.

**A criterion's first evaluation gets no grace period**, keyed on `last_verdict_at IS NULL`. One we have never seen pass has no earned pass for grace to hold.

**The `feedreliabilityseal` row is written for every evaluated feed**, qualifying or not — a "how long have we tracked this feed" criterion (#1782 Stable) measures against its `created_at`.

**`inactive` and `future` feeds are evaluated.** Excluding a feed freezes its rows rather than making it neutral. Only `deprecated`/`development` and non-`published` are excluded.

**Both seal tables are written with Core statements against `__table__`.** `feedreliabilityseal.feed_id` is both PK and FK to `feed(id)`, which sqlacodegen maps as joined-table inheritance, so persisting an ORM instance would try to insert a feed.

## Tests

105 tests across the package.

- `test_seal_evaluators.py` — the Official criterion and the base-class contract (including that an evaluator may not return `never_evaluated`). No database.
- `test_seal_state_machine.py` — grace expiry, streak reset, first-evaluation-no-grace, the `unknown`/`not_applicable` split, probation (open/serve/restart/expiry, day granularity), and probation suspending grace. No database.
- `test_seal_updater_db.py` — eligibility, context loading, grant/deny/revoke/recover, a criterion going dark (`unknown`) vs stopping applying (`not_applicable`), count balance. Test DB.
- `test_seal_end_to_end_db.py` — seeds feeds, invokes `tasks_executor` with a `flask.Request`, inspects both tables across two runs. Test DB.

The two DB suites need the local Postgres test database, like `test_check_gtfs_feed_availability_db.py`. CI provides it.

### DIFF
--- a/functions-python/tasks_executor/src/main.py
+++ b/functions-python/tasks_executor/src/main.py
@@ -261,8 +261,7 @@ tasks = {
     "update_seal_of_reliability": {
         "description": (
             "Evaluates the implemented Seal of Reliability criteria for every eligible "
-            "GTFS feed and updates sealcriterion and feedreliabilityseal. Only the "
-            "Official criterion is implemented so far (see #1784 and #1782 for the rest). "
+            "GTFS feed and updates sealcriterion and feedreliabilityseal. "
             "Reads the source tables and never modifies them. "
             "Parameters: dry_run (default true), stable_feed_ids (default null; when set, "
             "`evaluations` covers every criterion of those feeds), limit (default null), "
--- a/functions-python/tasks_executor/src/main.py
+++ b/functions-python/tasks_executor/src/main.py
@@ -264,12 +264,11 @@ tasks = {
     },
     "update_seal_of_reliability": {
         "description": (
-            "Evaluates the implemented Seal of Reliability criteria for every eligible "
-            "GTFS feed and updates sealcriterion and feedreliabilityseal. "
+            "Evaluates the implemented Seal of Reliability criteria for the requested "
+            "GTFS feeds and updates seal_criterion and feed_reliability_seal. "
             "Reads the source tables and never modifies them. "
-            "Parameters: dry_run (default true), stable_feed_ids (default null; when set, "
-            "`feeds` covers every one of those feeds), limit (default null), "
-            "criteria (default null "
+            "Parameters: stable_feed_ids (required, non-empty; the feeds to evaluate), "
+            "dry_run (default true), limit (default null), criteria (default null "
             "meaning every implemented criterion; a partial set skips the has_seal "
             "roll-up), batch_size (default 200), now (ISO timestamp, default current "
             "UTC time)."
--- a/functions-python/tasks_executor/src/main.py
+++ b/functions-python/tasks_executor/src/main.py
@@ -76,6 +76,9 @@ from tasks.notifications.dispatch_monitor import (
     notifications_dispatch_monitor_handler,
 )
 from tasks.changelog.backfill_changelog import backfill_changelog_handler
+from tasks.seal_of_reliability.update_seal_of_reliability import (
+    update_seal_of_reliability_handler,
+)
 
 init_logger()
 LIST_COMMAND: Final[str] = "list"
@@ -254,6 +257,21 @@ tasks = {
             "feeds_not_updated_days (default null)."
         ),
         "handler": backfill_changelog_handler,
+    },
+    "update_seal_of_reliability": {
+        "description": (
+            "Evaluates the implemented Seal of Reliability criteria for every eligible "
+            "GTFS feed and updates sealcriterion and feedreliabilityseal. Only the "
+            "Official criterion is implemented so far (see #1784 and #1782 for the rest). "
+            "Reads the source tables and never modifies them. "
+            "Parameters: dry_run (default true), stable_feed_ids (default null; when set, "
+            "`evaluations` covers every criterion of those feeds), limit (default null), "
+            "criteria (default null "
+            "meaning every implemented criterion; a partial set skips the has_seal "
+            "roll-up), batch_size (default 200), now (ISO timestamp, default current "
+            "UTC time)."
+        ),
+        "handler": update_seal_of_reliability_handler,
     },
 }
 
--- a/functions-python/tasks_executor/src/main.py
+++ b/functions-python/tasks_executor/src/main.py
@@ -264,7 +264,7 @@ tasks = {
             "GTFS feed and updates sealcriterion and feedreliabilityseal. "
             "Reads the source tables and never modifies them. "
             "Parameters: dry_run (default true), stable_feed_ids (default null; when set, "
-            "`evaluations` covers every criterion of those feeds), limit (default null), "
+            "`feeds` covers every one of those feeds), limit (default null), "
             "criteria (default null "
             "meaning every implemented criterion; a partial set skips the has_seal "
             "roll-up), batch_size (default 200), now (ISO timestamp, default current "
--- a/functions-python/tasks_executor/src/tasks/seal_of_reliability/context.py
+++ b/functions-python/tasks_executor/src/tasks/seal_of_reliability/context.py
@@ -1,0 +1,122 @@
+#
+#   MobilityData 2026
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Feed eligibility query and bulk loading of everything the evaluators need.
+
+The evaluators are pure functions over a `FeedSealContext`, so every DB read for a batch of
+feeds happens here, in a fixed number of queries regardless of batch size.
+
+Only Official is implemented, so the context currently carries just the feed row. Each new
+criterion adds the fields it needs here plus one bulk query to populate them: the latest
+dataset for Compliant and Fresh, the day's availability rows for Available, the full
+dataset coverage history for Fresh continuous coverage.
+"""
+
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Dict, Optional, Sequence
+
+from sqlalchemy.orm import Session
+
+from shared.database_gen.sqlacodegen_models import Feed, Gtfsfeed
+
+
+@dataclass
+class FeedSealContext:
+    """Everything the evaluators need for one feed.
+
+    Built by `build_contexts`. Evaluators read from this and never query.
+    """
+
+    feed_id: str
+    # The evaluation timestamp. Passed in rather than read from the clock so evaluators
+    # stay pure and a run can be replayed for any point in time.
+    now: datetime
+    stable_id: Optional[str] = None
+
+    # Feed-level flags
+    official: Optional[bool] = None
+
+
+def get_seal_feeds_query(
+    db_session: Session, stable_feed_ids: Optional[Sequence[str]] = None
+):
+    """Return a query for the feeds the seal applies to.
+
+    Eligibility is defined here and nowhere else, so a full run and a one-feed run exercise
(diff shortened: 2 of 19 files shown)


functions-python/tasks_executor/src/tasks/seal_of_reliability/context.py
```diff
@@ -0,0 +1,122 @@
...
+):
+    """Return a query for the feeds the seal applies to.
+
+    Eligibility is defined here and nowhere else, so a full run and a one-feed run exercise
+    the same predicate. `inactive` and `future` feeds are deliberately included: skipping a
+    feed does not make it neutral, it freezes its stored rows, and once Fresh (future
+    coverage) exists an inactive feed should fail it rather than keep displaying a seal.
+    """
+    query = db_session.query(Gtfsfeed).filter(
+        Feed.data_type == "gtfs",
```
cka-y: [nit] db_session.query(Gtfsfeed) already filters by `gtfs` data type.

liquibase/changes/feat_1783.sql
```diff
@@ -0,0 +1,86 @@
...
+CREATE TYPE seal_criterion_status AS ENUM (
+    'pass',
+    'fail',
+    'unknown',
+    'never_evaluated',
+    'not_applicable'
+);
+
+-- One row per feed; owns the overall seal outcome.
+CREATE TABLE FeedReliabilitySeal (
```
davidgamez: To have a generated code equal to `FeedReliabilitySeal`, we should rename the table to feed_reliability_seal.

liquibase/changes/feat_1783.sql
```diff
@@ -0,0 +1,86 @@
...
+    feed_id        VARCHAR(255) PRIMARY KEY REFERENCES Feed(id) ON DELETE CASCADE,
```
davidgamez: `feed_id` as both PK and FK to `Feed(id)` is the shape sqlacodegen reads as joined-table inheritance, so it generates `class Feedreliabilityseal(Feed)`.

```suggestion
CREATE TABLE FeedReliabilitySeal (
    id             UUID PRIMARY KEY DEFAULT gen_random_uuid(),
    feed_id        VARCHAR(255) NOT NULL UNIQUE REFERENCES Feed(id) ON DELETE CASCADE,
```

`UNIQUE` keeps one row per feed, and after `db-gen.sh` the model becomes a plain `Base` subclass with a 1:1 `feed` relationship. Would also need `Feed.feedreliabilityseal` added to `cascade_entities` in `database.py`, next to `Feed.sealcriteria`.

functions-python/tasks_executor/src/tasks/seal_of_reliability/context.py
```diff
@@ -0,0 +1,117 @@
...
+
+def get_seal_feeds_query(db_session: Session, stable_feed_ids: Sequence[str]):
+    """Return a query for the requested feeds that are eligible for the seal.
+
+    The task always runs against an explicit list of feeds. `deprecated`,
+    `development` and non-`published` feeds are dropped.
+    `inactive` and `future` feeds are deliberately kept.
+    """
+    return db_session.query(Gtfsfeed).filter(
+        Feed.data_type == "gtfs",
```
cka-y: [nit] db_session.query(Gtfsfeed) already filters by gtfs data type.

functions-python/tasks_executor/src/tasks/seal_of_reliability/update_seal_of_reliability.py
```diff
@@ -0,0 +1,80 @@
...
+def get_parameters(payload: dict):
+    """Read the task parameters from the payload, applying defaults."""
+    now = payload.get("now")
+    return (
+        payload.get("stable_feed_ids"),
+        payload.get("dry_run", True),
+        payload.get("limit", None),
+        payload.get("criteria", None),
+        payload.get("batch_size", DEFAULT_BATCH_SIZE),
+        datetime.fromisoformat(now) if now else None,
```
cka-y: [from Claude] The README documents `now` for replays/backfills; an operator naturally passes `"now": "2026-08-01"` or `"2026-08-01T00:00:00"` (no offset). `datetime.fromisoformat` returns a naive datetime, which flows unchanged through `update_seals` (`now = now or datetime.now(timezone.utc)` keeps it naive). DB timestamps (`probation_start`, `first_observed_failure_at`, ...) come back tz-aware from timestamptz columns, so `now - first_observed_failure_at < grace_period` and `now >= base.probation_start + probation_period` in `state_machine` raise `TypeError: can't subtract offset-naive and offset-aware datetimes` for any criterion with a grace or probation period. Official has neither so it doesn't trip today, but the state machine (grace/probation) shipped in this PR, and the naive value is also written into timestamptz columns ambiguously. Normalize to UTC (assume/attach tzinfo) in `get_parameters`.

functions-python/tasks_executor/src/main.py
```diff
@@ -259,6 +262,20 @@
         ),
         "handler": backfill_changelog_handler,
     },
+    "update_seal_of_reliability": {
+        "description": (
+            "Evaluates the implemented Seal of Reliability criteria for every eligible "
+            "GTFS feed and updates sealcriterion and feedreliabilityseal. "
+            "Reads the source tables and never modifies them. "
+            "Parameters: dry_run (default true), stable_feed_ids (default null; when set, "
```
cka-y: should we change this to required?
```suggestion
            "Parameters: dry_run (default true), stable_feed_ids (required, "
```

cka-y (approved):
lgtm added minor comments 🎖️

liquibase/changes/feat_1783.sql
jcpitre: Makes sense.
We used different ways of naming tables in the past, including camelcase (or PascalCase?) like done here. I guess I chose to replicate the wrong one.

liquibase/changes/feat_1783.sql
jcpitre: Yeah, I saw that. I agree, but is there a problem with having Feedreliabilityseal as a child of Feed? Apart from the semantics that tells us it should be completely separate.

liquibase/changes/feat_1783.sql
jcpitre: Done

liquibase/changes/feat_1783.sql
jcpitre: Done

davidgamez: > Yeah, I saw that. I agree, but is there a problem with having Feedreliabilityseal as a child of Feed? Apart from the semantics that tells us it should be completely separate.

It's inconsistent what it represents. Having it as a child means Feedreliabilityseal is a "Feed," like GBFS or GTFSRealtime.

functions-python/tasks_executor/src/tasks/seal_of_reliability/update_seal_of_reliability.py
jcpitre: Wow!

davidgamez (approved):

functions-python/tasks_executor/src/main.py
jcpitre: Done

functions-python/tasks_executor/src/tasks/seal_of_reliability/update_seal_of_reliability.py
jcpitre: Done

functions-python/tasks_executor/src/tasks/seal_of_reliability/context.py
jcpitre: Done

functions-python/tasks_executor/src/tasks/seal_of_reliability/update_seal_of_reliability.py
jcpitre: @cka-y are you ok with the timezone change?
Because if you are it's merge time!

PR #1796, opened 2026-08-06
feat: extract `feed_info.txt` content

### LINKED ISSUE #1775: Expose feed start and end date from validation report in our DB and/or API
For calculating continuous service windows, we need to be evaluating both the feed_info.txt start and end dates and the calendar.txt and calendar_dates.txt start and end dates. [See more details here.](https://docs.google.com/document/d/16_B5Bfw1LIQ2VbitlUny0PBERcMOTvXa_2D0LHS30D0/edit?tab=t.0#heading=h.8i4ko6yr1kls)

That means we need to be exposing the feed start and end dates from the validation report in our database so we can use them for our calculation. 

To be discussed if we need them both in the database and the public API or not.

We'll also need to backfill these values for past datasets.

cka-y:
**Summary:**
Closes #1775
Creates the architecture to extract file content from GTFS files, plus the first extractor: `feed_info.txt`. A new `gtfs-file-data-extractor` function receives one file per invocation and delegates to an extractor resolved by file name, so supporting a new file means adding a `FileDataExtractor` subclass, registering it in `extractors/registry.py`, and adding the file name to `EXTRACTABLE_FILES` in `pipeline_tasks.py`.
Adds a `feedinfo` table with one row per distinct `feed_info.txt` content, keyed by its content hash and shared by every dataset with the same content (`gtfsdataset.feed_info_id` -> `feedinfo.id`, one `feedinfo` to many datasets). No API spec changes; the only existing-table change is the new nullable `gtfsdataset.feed_info_id` (`ON DELETE SET NULL`).

**Expected behavior:**
After the daily batch processing (`batch-process-dataset`), we have a list of files registered for content extraction. For each file we trigger the extraction providing the file url and dataset and feed information.
A task is enqueued for every extractable file present, changed or not, since an unchanged file may never have been extracted before. The extractor links the dataset to the row for that content hash when one already exists (no download), otherwise it downloads and extracts a new one. Failures return 200 with the error in the body, so Cloud Tasks does not retry what can never succeed; every path is idempotent. Existing datasets are not backfilled (follow-up ticket).

**Deploy note:** `feat_1775.sql` now drops and recreates `feedinfo` (an earlier revision created it one row per dataset), so environments that ran that revision need its `databasechangelog` row deleted first, or Liquibase aborts on a checksum mismatch. Never applied to prod.

**Testing tips:**
Function can be tested by running the batch processing function or by calling the newly created `gtfs-file-data-extractor` function directly through HTTP after building this branch on DEV/QA. E.g. of running it in QA:

```bash
curl -X POST "https://northamerica-northeast1-mobility-feeds-qa.cloudfunctions.net/gtfs-file-data-extractor" \
-H "Content-Type: application/json" \
-d '{
  "stable_id": "mdb-1884",
  "dataset_id": "mdb-1884-202508220038",
  "file_name": "feed_info.txt",
  "file_url": "https://storage.googleapis.com/mobilitydata-datasets-prod/mdb-1884/mdb-1884-202508220038/extracted/feed_info.txt"
}'
```

For the sharing path, run it on two datasets of the same feed with an identical `feed_info.txt`: both end up on the same `feedinfo.id`.

Note for deploying: `api-deployer` creates the function, `datasets-batch-deployer` creates the task queue. Run `api-deployer` first, otherwise tasks are enqueued against a function that doesn't exist yet.

### DIFF
--- a/functions-python/batch_process_dataset/src/pipeline_tasks.py
+++ b/functions-python/batch_process_dataset/src/pipeline_tasks.py
@@ -146,11 +146,16 @@ def create_pipeline_tasks(dataset: Gtfsdataset, db_session: Session) -> None:
             stable_id, dataset_stable_id, stops_url
         )
 
-    # Create GTFS file data extraction tasks for changed, extractable files.
+    # Create GTFS file data extraction tasks for every extractable file present in
+    # the dataset, whether or not it changed: each dataset needs its own reference
+    # to the extracted data, and an unchanged file may never have been extracted
+    # before. Extracted data is shared by file content hash, so for an unchanged
+    # file the extractor just links the dataset to the existing data instead of
+    # downloading the file again.
     files_by_name = {file.file_name: file for file in gtfs_files}
     for file_name in EXTRACTABLE_FILES:
         gtfs_file = files_by_name.get(file_name)
-        if gtfs_file and gtfs_file.hosted_url and file_name in changed_files:
+        if gtfs_file and gtfs_file.hosted_url:
             create_http_gtfs_file_data_extractor_task(
                 stable_id, dataset_stable_id, file_name, gtfs_file.hosted_url
             )
--- a/functions-python/batch_process_dataset/src/pipeline_tasks.py
+++ b/functions-python/batch_process_dataset/src/pipeline_tasks.py
@@ -147,11 +147,7 @@ def create_pipeline_tasks(dataset: Gtfsdataset, db_session: Session) -> None:
         )
 
     # Create GTFS file data extraction tasks for every extractable file present in
-    # the dataset, whether or not it changed: each dataset needs its own reference
-    # to the extracted data, and an unchanged file may never have been extracted
-    # before. Extracted data is shared by file content hash, so for an unchanged
-    # file the extractor just links the dataset to the existing data instead of
-    # downloading the file again.
+    # the dataset, whether it changed or not
     files_by_name = {file.file_name: file for file in gtfs_files}
     for file_name in EXTRACTABLE_FILES:
         gtfs_file = files_by_name.get(file_name)
--- a/functions-python/batch_process_dataset/src/pipeline_tasks.py
+++ b/functions-python/batch_process_dataset/src/pipeline_tasks.py
@@ -14,6 +14,10 @@ from shared.helpers.utils import (
     create_http_gtfs_datasets_comparer_task,
 )
 
+# GTFS files that have a registered extractor in the gtfs_file_data_extractor
+# function. Keep in sync with that function's src/extractors/registry.py.
+EXTRACTABLE_FILES = {"feed_info.txt"}
+
 
 def create_http_reverse_geolocation_processor_task(
     stable_id: str,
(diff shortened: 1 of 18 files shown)


functions-python/batch_process_dataset/src/pipeline_tasks.py
```diff
@@ -109,6 +146,15 @@ def create_pipeline_tasks(dataset: Gtfsdataset, db_session: Session) -> None:
             stable_id, dataset_stable_id, stops_url
         )
 
+    # Create GTFS file data extraction tasks for changed, extractable files.
+    files_by_name = {file.file_name: file for file in gtfs_files}
```
jcpitre, line 151: I have questions about this loop. 
It is possible that a dataset has the same feed_info.txt as it's predecessor. For example a feed_info.txt that does not specify the dates, that are optional, might not have changed.
In that case we do not extract feed_info.txt again. right?
That means the feedinfo table will not have an entry for the second dataset. Correct me if I'm wrong.
How will we access feed_info.txt data for the second dataset, even if it's only to verify that it has no dates? Plus we might be interested in other feed_info.txt data in the future.
Maybe it's not worth it to not extract feed_info.txt if it has not changed?

functions-python/gtfs_file_data_extractor/src/processor.py
```diff
@@ -0,0 +1,101 @@
...
+        db_session.commit()
+    except Exception as error:
+        db_session.rollback()
+        logging.error(
+            "Error extracting data from %s for dataset %s: %s",
+            file_name,
+            dataset_id,
+            error,
+        )
+        return f"Error extracting data from {file_name}: {error}", ERROR_STATUS_CODE
```
jcpitre: In pmtiles and others we return 200 each time we don't want the queue to do a retry, and use the message to signify the failure.
I'm not sure but I think an error 500 leads to a retry. If not ignore this comment.
You should consider doing the same for any error that you know will give the same result in a retry. 
And reserve 500 for errors that are retryable. Although i'm not sure you can discriminate here.
(Actually sometimes we should return 4XX. sometime 5XX, but that's another debate)

functions-python/batch_process_dataset/src/pipeline_tasks.py
cka-y: your understanding is almost fully correct.  we extract the `feed_info.` data if `feed_info.txt` hash has changed, so that does not limit it to data we save (the future case is covered).

Are you suggesting a different behaviour?

jcpitre: Let me rephrase. If a dataset has exactly the same feed_info.txt as its predecessor, then there will not be an entry in the feedinfo table for this dataset.
Even if in fact this dataset _does_ have a feed_info.txt.

E.G.: I could easily see someone making corrections to a feed (of say a route name), but the dates in feed_info.txt (and all of feed_info.txt) remain the same.
Then there will not be a feedinfo entry for that latest dataset.

cka-y: Are you suggesting a different behaviour?

jcpitre: But you agree there's a problem?
Because I could we wrong. It happened in the past :smile:
You could either always extract feed_info.txt, even if it did not change, 
or duplicate the `feedinfo` record of the predecessor but just change the `gtfs_dataset_id` to the id of the latest dataset.

liquibase/changes/feat_1775.sql
```diff
@@ -0,0 +1,26 @@
+-- Store parsed feed_info.txt fields per GTFS dataset (issue #1775).
+-- One row per dataset; populated by the gtfs_file_data_extractor cloud function.
+CREATE TABLE IF NOT EXISTS feedinfo (
```
davidgamez: Since feed_info is unlikely to change, getting the feed_info for a "gtfs_feed" would require constructing a query using the datasets. I'm proposing to create a relationship between the dataset that "contains" the feed_info and the feed; the same pattern that we have for

jcpitre (commented):
LGTM!

jcpitre (approved):

davidgamez (approved):

PR #1797, opened 2026-08-06
feat: odpt integration

### LINKED ISSUE #1391: Set up automatic import of ODPT Feeds
Automatically import feeds in ODPT Center with CC licence. (Do not add feeds with other licences)

Static[CC BY 4.0]: https://ckan.odpt.org/dataset/?res_format=GTFS%2FGTFS-JP&license_id=CC-BY-4.0
(40 feeds)

Static[CC0 1.0]: https://ckan.odpt.org/dataset/?res_format=GTFS%2FGTFS-JP&license_id=CC0-1.0
(3 feeds)

Realtime[CC BY 4.0]: https://ckan.odpt.org/dataset/?res_format=Protocol+Buffers&license_id=CC-BY-4.0
(50 feeds)

Please review the requirements for republishing: https://developer.odpt.org/top 

API documentation is here: https://developer.odpt.org/documents

ianktc:
**Summary:**

[odpt_script](https://github.com/user-attachments/files/30803053/odpt_gtfs_list.5.ipynb) as provided directly from ODPT

This PR adds a new odpt_import task to the tasks_executor Cloud Function that imports Japanese transit feeds from the Open Data Platform for Transportation (ODPT). It queries ODPT's members-portal metadata API for openly-licensed (CC BY 4.0 and CC0) GTFS datasets, creates/updates the corresponding Gtfsfeed and Gtfsrealtimefeed (trip updates, vehicle positions, service alerts) records, tags each with an odpt external ID and a Japan-level Location, and links related GTFS-RT feeds back to their parent schedule feed. Closes MobilityData/mobility-database-catalogs#1391. A new monthly Cloud Scheduler job (odpt_import_schedule, 5th of the month) is added, alongside the existing JBDA/TDG import schedules.

**Expected behavior:** 

- Running the task with dry_run: true (default) fetches and diffs against the DB but performs no writes — all staged session changes are explicitly rolled back at the end.
- Running with dry_run: false creates new GTFS/GTFS-RT feeds for previously-unseen ODPT org/dataset pairs, updates existing ones only when a tracked field (name, provider, producer URL, license URL, or an RT endpoint URL) actually changed, and leaves unchanged feeds untouched.
- New/changed feeds trigger a dataset-download Pub/Sub message; changed feeds also enqueue a website cache-revalidation task.
- If a schedule feed's producer URL changes in place, a feed.url_updated notification event is emitted.
- Changes are committed in batches (COMMIT_BATCH_SIZE, default 5) rather than one giant transaction, with per-batch IntegrityError handling that rolls back just that batch.

**Testing tips:**

- Run scripts/api-tests.sh --folder functions-python/tasks_executor — includes the new test_odpt_import.py covering: full create path (schedule + 3 RT feeds + location + external ID), the dry-run-must-not-write regression test, and graceful handling of an upstream HTTP failure.
- To exercise locally end-to-end, invoke the tasks_executor function with {"task": "odpt_import", "payload": {"dry_run": true}} first and inspect the returned summary (created_gtfs, updated_gtfs, created_gtfs_rt, updated_gtfs_rt, linked_refs, total_processed_items) before flipping to dry_run: false against a local DB.


ianktc:
I found the same issue with @with_db_session annotation committing on dry_run=True without an explicit rollback. This is addressed by adding an explicit rollback. `if not dry_run then commit else rollback.` I believe the same issue exists in tdg and jbda imports.

Explanation:

1. _import_odpt(dry_run=dry_run) is called with no db_session kwarg (line 88 of import_odpt_handler).
2. with_db_session's wrapper sees kwargs.get("db_session") is None, so it creates a Database instance and opens db.start_db_session() itself, injecting the new session into kwargs["db_session"].
3. start_db_session (lines 242–250) is a plain try/except: yield session (runs _import_odpt's whole body) → session.commit() if nothing raised → session.rollback() only if an exception propagated out.
4. _import_odpt returns a normal dict in the dry-run case (it never raises an exception) so branch is always "commit," never "rollback."

functions-python/tasks_executor/src/tasks/data_import/odpt/import_odpt_feeds.py
```diff
@@ -0,0 +1,565 @@
+#!/usr/bin/env python3
```
cka-y, line 1: there's no deprecation path: if a dataset is withdrawn, or its license changes from CC BY/CC0 to something closed, it drops out of the query and the feed stays status="active", operational_status="published" forever.

functions-python/tasks_executor/src/tasks/data_import/odpt/import_odpt_feeds.py
```diff
@@ -0,0 +1,565 @@
...
+    logger.debug(
+        "Updating common fields for feed id=%s stable_id=%s",
+        getattr(feed, "id", None),
+        getattr(feed, "stable_id", None),
+    )
+    feed.feed_name = item.get("dataset_name_ja") or item.get("dataset_name_en")
+    feed.provider = item.get("org_name_ja") or item.get("org_name_en")
+    feed.producer_url = producer_url
+    feed.license_url = _get_license_url(item.get("license_type"))
+    feed.status = "active"
```
cka-y: [suggestion] status is computed every night based on the service window. from my understanding this import will unconditionally overwrite the value. so an inactive feed could be set as active. not a big deal because the daily status computation will fix it overnight but worth flagging/fixing.

functions-python/tasks_executor/src/tasks/data_import/odpt/import_odpt_feeds.py
```diff
@@ -0,0 +1,565 @@
...
+        for feed in feeds_to_publish:
+            trigger_dataset_download(feed, execution_id)
+        if changed_feed_stable_ids:
+            try:
+                create_web_revalidation_task(changed_feed_stable_ids)
+            except Exception as e:
+                logger.warning("Failed to enqueue revalidation tasks: %s", e)
+    except IntegrityError:
+        db_session.rollback()
+        logger.exception("Commit failed with IntegrityError; rolled back")
```
cka-y: [nit] this is an exact duplication from the jbda import. suggesting to move to common file like `data_import_utils.py` to avoid code duplication.

functions-python/tasks_executor/src/tasks/data_import/odpt/import_odpt_feeds.py
```diff
@@ -0,0 +1,565 @@
...
+                    old_url=old_url,
+                    new_url=new_url,
+                    source="odpt_import",
+                )
+
+    # Apply schedule fields
+    _update_common_feed_fields(gtfs_feed, item, producer_url)
+
+    # Location (append only if empty)
+    location = _get_or_create_location(db_session)
```
cka-y: [nit] no need to do a db query per feed since they all share the same location (Japan) - could be a variable outside of the for loop.

cka-y (commented):
[test notes] tested in dev - got 89 new feeds + scheduler success run + pipeline successfully wired for odpt feeds

ianktc:
### Regarding Deprecation Path
How are feeds deprecated if they disappear from source API?

- `deprecate_stale_feeds` located in `data_import_utils.py` compares a set of processed_stable_ids (from the source API) with the feeds in our db with the target prefix (in this case odpt)
    - it will deprecate any feeds that were not seen in the source API (because they were removed or because their license changed) and mark them to unpublished so that we don’t display feeds that were removed in the portal

How are feeds reactivated (to active and published) if they reappear in the source API?

- `_process_feed` will check for any schedule feeds that need reactivation and `_extract_db_rt_map` will check for any realtime feeds related to these schedule feeds that also need reactivation
    - reactivation will occur in `_update_common_feed_fields`

The _import_odpt handler will always sweep for deprecated feeds even on dry_run=True so that it can report how many feeds *would* be deprecated.

cka-y:
[question] any reason we're only including feeds with CC BY/CC0 licenses?

ianktc:
> [question] any reason we're only including feeds with CC BY/CC0 licenses?

I believe those are the only licenses that ODPT said we can publish cc. @tzujenchanmbd

tzujenchanmbd:
These are the licenses in the ODPT portal that currently permit the redistribution of data.
Other licenses, such as the "Public Transportation Open Data Basic License," prohibit data redistribution.

cka-y (approved):
LGTM!

Issue #1803, opened 2026-08-19
v2: Seal of Reliability: Re-evaluate a feed when a past UNKNOWN becomes known


jcpitre:
A criterion can come out UNKNOWN on a given day because an input was missing — e.g.
Compliant has no validation report yet for the latest dataset. When that input arrives
later (the report lands after the fact), the stored seal state was computed without it and
may now be wrong. We need to re-run the evaluation so the feed's seal reflects it.

## How

Reuse the #1763 backfill: cold-start the feed and march to today, saving only today's
state — just triggered for a single feed when a late input is detected, rather than as a
bulk run. Because the state is path-dependent (see #1763), the whole window is re-derived
rather than a single day patched.

## Note
**The following is now covered by #1809**
A daily state snapshot (one `seal_criterion` row per feed, criterion and day) would:
- let a correction resume from the day before the change instead of cold-starting the whole
  window;
- preserve each criterion's past values — a real historical record — making re-evaluation
  more precise, since Official and Stable could be judged against their actual past values
  (only back to when snapshotting began, though; earlier days still fall back to assumed
  values — likely the earliest one we recorded);
- cost DB space (one row per feed, per criterion, per day), which would need to be
  evaluated but is not expected to be huge.

- We should consider the possibility of an audit trail for that, so we can explain to the user why a seal changed after a reevaluation.

PR #1810, opened 2026-08-22
Added seal_criterion_snapshot table.

### LINKED ISSUE #1809: Seal of Reliability: Persist the daily state of the criteria
In #1803 we proposed to have the `seal_criterion` DB table record that daily state of each criterion, instead of just keeping "yesterday's" state:

> A daily state snapshot (one `seal_criterion` row per feed, criterion and day) would:
> 
> * let a correction resume from the day before the change instead of cold-starting the whole
>   window;
> * preserve each criterion's past values — a real historical record — making re-evaluation
>   more precise, since Official and Stable could be judged against their actual past values
>   (only back to when snapshotting began, though; earlier days still fall back to assumed
>   values — likely the earliest one we recorded);
> * cost DB space (one row per feed, per criterion, per day), which would need to be
>   evaluated but is not expected to be huge.

Let's implement this immediately.

jcpitre:
## From our AI friend:
`seal_criterion` holds only the current state and is overwritten on every run, so past values
are lost and a retroactive correction has to cold-start a feed's whole window.

Adds `seal_criterion_snapshot`: one row per feed, criterion and evaluated day, written by the
nightly job from the same state and in the same transaction as the `seal_criterion` write.
Keyed on `(feed_id, criterion, snapshot_date)` with the day as a `DATE`, so re-running or
replaying a day retakes that snapshot rather than adding a row. Gaps are normal — a reader
takes the latest snapshot on or before the day it asks about.

`seal_criterion` is unchanged, so nothing reading it today has to move. The job writes the
snapshots but never reads them; #1803 will, to resume a correction from the day before the
one it is changing.

### DIFF
--- a/api/src/shared/database/database.py
+++ b/api/src/shared/database/database.py
@@ -78,6 +78,7 @@ cascade_entities = {
         Feed.redirectingids_,  # redirectingid_target_id_fkey
         Feed.feed_license_changes,
         Feed.seal_criteria,  # seal_criterion_feed_id_fkey
+        Feed.seal_criterion_snapshots,  # seal_criterion_snapshot_feed_id_fkey
         Feed.feed_reliability_seal,  # feed_reliability_seal_feed_id_fkey
     ],
     Gtfsfeed: [Gtfsfeed.gtfs_dataset_changelogs],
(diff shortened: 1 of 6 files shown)


davidgamez (approved):
LGTM

PR #1811, opened 2026-08-22
Fix GTFS-RT location inheritance from static feeds

mackenziereading19:
## Summary

Fixes GTFS-RT feeds remaining locationless when they are imported after a referenced static GTFS feed already has location metadata.

During catalogue population, once `static_reference` values are resolved, a GTFS-RT feed with no location metadata of its own now inherits the locations already attached to its matched static GTFS feed(s).

Existing GTFS-RT locations are preserved. If multiple static feeds are referenced, their locations are combined and deduplicated by location ID.

This addresses MobilityData/mobility-database-catalogs#1567, where recently imported STM and Metra GTFS-RT feeds had valid static feed references but empty `locations`.

## Tests

Added regression coverage for:

- a new locationless GTFS-RT feed inheriting an existing static feed location;
- preserving an existing GTFS-RT location rather than overwriting it;
- combining locations from multiple static feed references.

Local verification:

- targeted regression tests: 3 passed;
- full `api/tests/integration/populate_tests/test_populate.py`: 20 passed;
- `functions-python/reverse_geolocation/tests/test_reverse_geolocation_processor.py`: 19 passed.

The reverse-geolocation suite was run using the repository's standard `function-python-setup.sh` shared/test-shared symlink setup.


davidgamez (approved):
LGTM!
@mackenziereading19  Thanks for your contribution

PR #1817, opened 2026-08-26
feat: evaluate fresh + stable criteria

cka-y:
**Summary:**

Implements two of the four criteria from #1784 — **Stable** and **Fresh (future coverage)** — following the algorithm spike in [[#1761](https://github.com/MobilityData/mobility-feed-api/issues/1761#issuecomment-4987530241)](https://github.com/MobilityData/mobility-feed-api/issues/1761#issuecomment-4987530241). They join `Official` (#1783) in the evaluator registry the nightly job iterates, so the `has_seal` roll-up now weighs three criteria instead of one. Available and Compliant remain open under #1784; Fresh / continuous coverage stays with #1782.

Two deliberate divergences from the spike, both worth reviewer attention:

- **Stable reads `feed.created_at`, not `feed_reliability_seal.created_at`.** The question is whether we hold six months of history for the feed, so the clock is the feed's own age. Reading the seal row would have meant no feed could qualify until 180 days after the job's first run, and would have made Stable unreplayable, since that row's `created_at` is stamped by the database clock rather than the run's `now`.
- **"Latest dataset" is resolved as of the run's `now`**, via `downloaded_at`, rather than from the `gtfsfeed.latest_dataset_id` pointer. That pointer always names the newest dataset that exists *today*, so a replay reading it would judge a past day against data published after it.

This PR also centralizes the seal's policy surface, which was duplicated between the job and the read API. `api/src/shared/common/seal_criteria.py` is now the only place any policy value is defined — the criterion names, `CriterionStatus`, `CriterionPhase`, the grace and probation maps, and the two check-internal windows (`TRACKING_PERIOD`, `FUTURE_COVERAGE_HORIZON`). Evaluators no longer declare their own windows: `CriterionEvaluator` resolves both from `self.name` via `grace_period_for()` / `probation_period_for()`, so a criterion cannot debounce one way for the job and another for the API. The API side drops its local `STATUS_PASS`/`STATUS_FAIL` string constants for the shared enum, and `tasks/seal_of_reliability/criteria.py` is deleted rather than left as a re-export shim.

No schema change: `seal_criterion_name` already declared all six values. No change to any public endpoint or response shape.

**Expected behavior:** 

**Stable** — passes when `feed.created_at <= now - 180 days` and `is_producer_url_unstable IS NOT TRUE`. No grace period and no probation, so it is a point-in-time check like Official: it fails the day the unstable flag is set and clears the day it is cleared. It never returns `unknown` — both inputs are columns on the feed row. A producer URL change creates a new feed with its own `created_at`, so the six months restart with it. Feeds already in the catalog for six months qualify on the very first run.

**Fresh (future coverage)** — passes when the latest dataset as of `now` has `service_date_range_end >= now + 7 days`. 14-day grace period, 180-day probation. Returns `not_applicable` for a seasonal feed, which withdraws the criterion from the roll-up rather than failing it. Returns `unknown` in two distinct cases, reported separately: the feed had no dataset at all as of the run, or it had one whose `service_date_range_end` was never extracted. Neither is read as a failure — an upstream gap must not deny seals across the catalog.

A criterion that has never produced a verdict gets no grace period, so a feed whose first Fresh verdict is a failure is a confirmed failure that same day rather than passing for two weeks on evidence we do not have.


api/src/shared/common/seal_criteria.py
```diff
@@ -63,6 +111,33 @@ class SealCriterionName(str, Enum):
 )
 
 
+# Stable: how long we must have been tracking a feed - measured from its
+# `feed_reliability_seal.created_at` - before it can be called stable.
```
jcpitre: We're using feed.created_at, no?
```suggestion
# `feed.created_at` - before it can be called stable.
```

api/src/shared/common/seal_criteria.py
```diff
@@ -30,6 +30,54 @@ class SealCriterionName(str, Enum):
...
+    NEVER_EVALUATED = "never_evaluated"
+    NOT_APPLICABLE = "not_applicable"
+
+    @property
+    def is_verdict(self) -> bool:
+        """True for PASS and FAIL - the two values that mean the check actually answered."""
+        return self in (CriterionStatus.PASS, CriterionStatus.FAIL)
+
+
+class CriterionPhase(str, Enum):
```
davidgamez, line 64: 👍

api/src/shared/db_models/reliability_criterion_impl.py
```diff
@@ -87,8 +84,8 @@ def from_orm(cls, criterion_row: SealCriterionOrm | None) -> ReliabilityCriterio
         # nothing left for grace to protect.
         in_grace_period = (
             grace_period is not None
-            and status == STATUS_FAIL
-            and criterion_row.confirmed_status == STATUS_PASS
+            and status == CriterionStatus.FAIL
```
jcpitre, line 87: That's a problem that was there before your PR.
The problem is UNKNOWN. We decided that UNKNOWN would continue the last verdict (pass or fail). But unknown will make this test fail, getting out of the grace period.
The algorithm tells us to use this test instead: 
```
criterion_row.first_observed_failure_at is not None
```

cka-y: yes this will be covered outside of the scope of this pr

api/src/shared/common/seal_criteria.py
cka-y: yes you're right. fixed ✅

davidgamez (approved):

PR #1806, opened 2026-08-19
feat: Seal of reliability orchestrator

davidgamez:
**Summary:**

Summary

Implements #1800: a Cloud Tasks fan-out that runs update_seal_of_reliability across every eligible feed in the catalog nightly, without hitting tasks_executor's own invocation timeout as the catalog grows.

- seal_orchestrator (producer) — resolves every seal-eligible GTFS feed (same eligibility rule update_seals already applies), splits stable_ids into batches (default 250 feeds/batch, ~16 batches at current catalog size), registers the run in TaskExecutionTracker, and enqueues one worker Cloud Task per batch plus a single monitor task.
- seal_orchestrator_worker (one per batch) — calls update_seals() for its slice and reports success/failure back to the tracker, storing its result as tracked metadata.
- seal_orchestrator_monitor (barrier, one per run) — polls via native Cloud Tasks retry (503/TaskInProgressError) until every batch reports, aggregates the per-batch results, and finalizes the run. A hard deadline_seconds (default 1h) guarantees it can never poll forever if a worker dies mid-run — past the deadline it force-settles the run as failed rather than waiting indefinitely.
- Any batch failure, or any batch left unaccounted for at the deadline, marks the whole run failed — this is stricter than the existing notification-dispatch monitor, since an incomplete seal run should be a visible failure, not a silent partial success.

### Supporting changes

- TaskExecutionTracker.mark_completed gained an optional metadata param (backward compatible) so a worker can stash its result for the monitor to aggregate later, without a dedicated results table.
- get_seal_feeds_query's stable_feed_ids is now optional, so the same eligibility predicate can enumerate the whole catalog (previously it only ever narrowed an explicit list).
- Terraform: two new Cloud Tasks queues (worker + monitor, mirroring the existing notification-dispatch queues) and a nightly Cloud Scheduler job (04:00 UTC, paused outside prod).
- README section documenting all three tasks' payloads/response fields.

**Expected behavior:** 

Explain and/or show screenshots for how you expect the pull request to work in your testing (in case other devices exhibit different behavior).

**Testing tips:**

Provide tips, procedures and sample files on how to test the feature.
Testers are invited to follow the tips AND to try anything they deem relevant outside the bounds of the testing tips. 

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Run the unit tests with `./scripts/api-tests.sh` to make sure you didn't break anything
- [ ] Add or update any needed documentation to the repo
[...]

PR #1818, opened 2026-08-26
feat: Create backfill function for seal of reliability

### LINKED ISSUE #1763: Create backfill function for Seal of Reliability
When we first turn on the Seal of Reliability, feeds have no stored state to carry
forward. Create a backfill function that establishes the current seal state for every
GTFS feed that doesn't have one yet, so the daily check (#1764) has a "yesterday" to
step from.

## How

For each feed, cold-start it **12 months ago** and run the nightly evaluation (the
algorithm in #1761) day by day, forward to yesterday. Marching forward builds up the
path-dependent state (grace-period streaks, probation) so the final state comes out right.

The cold start assumes an empty prior state going into the first day — no ongoing failure
streak, no probation — which may not match reality, so early days can be wrong. Those
errors decay as the simulation runs. 12 months is chosen as roughly twice the memory
horizon (probation is ~6 months), which should be enough for the cold start to work itself
out before we reach the recent days that actually determine the final state.

Only the final state (yesterday) is written to `seal_criterion` and `feed_reliability_seal`
— the state the daily check then applies today's evaluation to. The intermediate days are
computed in memory and discarded — we do not store a row per day.

## Notes

- Official and Stable have no historical record, so the backfill can only evaluate them
  against their current values. They aren't expected to change much (maybe once in a feed's
  life), so current values are rarely wrong anyway; and neither has a grace period or
  probation, so a wrong value on a past day does not propagate forward — acceptable.
- Revisiting a day whose result was UNKNOWN and only became known later (e.g. a late
  validation report) is a separate concern handled by #1803. It would be nice if the
  backfill code here were written so it can serve that case too.

jcpitre:
Closes #1763 
## Summary

Added some backfill functions in tasks_executor.
The orchestration is similar to what is done in #1806

Added a simulation mode that allows testing different scenarios.

### What this adds

`backfill_seal_of_reliability` (#1763): gives feeds with no seal state a starting one for the
nightly job (#1761) to step from. Each feed is cold-started at
`march_start = max(start_date, created_at)`, the nightly evaluation is replayed forward a day at
a time to `end_date`, and only the final day is written.

Plus the Cloud Tasks fan-out across the catalog (`seal_backfill_orchestrator`,
`seal_backfill_worker`), sharing `fanout.py` and the nightly monitor, and README docs for all of
it.

### Algorithm fix (commit f689569a)

"Corrected a problem with UNKNOWN and the grace period" touches shared `transition` code, so it
affects the nightly job too, not just the backfill.

Grace expiry was only evaluated on a day the criterion was observed failing. A streak that
reached its deadline on an UNKNOWN day escaped confirmation entirely, and the next passing day
erased it — fourteen failing days leaving `last_confirmed_failure_at` NULL. Expiry is now a
property of the streak and the clock.

### Simulate and trace

`simulate` forces observed statuses per criterion (a `default` baseline plus named day
exceptions, offsets from each feed's own `march_start`). **It never writes** — a forced verdict
is indistinguishable from an earned one — so `trace` is the substitute, with two tests pinning
that it carries every field a snapshot would, day for day. `collapse_trace` defaults to `true`;
`false` gives one row per marched day.

### Tests run

`324 passed`. Beyond that, a local plan runs 58 assertions end to end over four feeds of
different ages: window resolution, the first write, re-runs and partial rewrites, resume,
grace/probation under `fresh_coverage`'s own 14/180 policy, the roll-up, and eight guardrails.
Smoke-tested on DEV, where four real feeds came out with march lengths of 366, 360, 228 and 4
days against one window — the per-feed clamping is the point.

### Worth a look during review

1. `official` has no history, so a currently-failing feed gets `first_observed_failure_at` at
   the **window edge** — specific, and unsupported by the data. `resume_from_snapshot` is the
   mitigation. `stable`'s equivalent timestamp is genuinely derivable, and correct.
2. A feed younger than `fresh_coverage`'s 14-day grace can never confirm a failure, so it is
   handed over holding an open streak.
3. Probation cannot expire in a window near its own 180-day horizon — the argument for
   `days_back = 365`.
4. A re-march replaces rather than accumulates: failure history the previous run recorded is
   lost if the new march never sees it.
5. Probation suspends grace (intended): a failing day during probation confirms immediately and
   re-stamps `probation_start`.

### Open item

Seven seal tasks could be four: both workers are their engine plus tracker bookkeeping, and the
two producers differ only in eligibility, window and three defaults. Follow-up, not this PR.


jcpitre:
@cka-y @davidgamez Modifications after our discussion:

1. **Inverted the march loops** — feeds outside, days inside. Same state produced; enables the rest.

2. **A backfill can be restarted after an interruption.** The atomic unit is the feed: its
   criteria, seal row and snapshots commit together once its march finishes, so an interrupted
   run leaves every finished feed durable and no feed half-written. Re-run the command to
   finish it.

   `only_missing` counts a feed as done only when it holds a `seal_criterion` row for every
   criterion the run evaluates — so a feed written by an earlier run over fewer criteria is marched again rather than skipped.

3. **Any unknow/never_evaluated observed statuses are re-evaluated** If the snapshot contains pass/fail/not_applicable, we don't revaluate and use the stored value.
   Everything else (`confirmed_status`, `probation_start`, …) is always recomputed.

4. **`snapshot_mode` now defaults to `all`** — storing only the endpoint threw away the history
   the march exists to build. ~2,200 rows per feed per year.

5. **The API `ReliabilityCriterion.status` now serves `confirmed_status`**

functions-python/tasks_executor/src/tasks/seal_of_reliability/context.py
```diff
@@ -84,34 +97,54 @@ class ClosestDataset:
...
     stable_id: Optional[str] = None
 
-    # Feed-level flags
+    # Kind 1: read straight off the feed row.
     official: Optional[bool] = None
     is_producer_url_unstable: Optional[bool] = None
     seasonal: Optional[bool] = None
 
-    # Stable: when the feed was first added to the database.
+    # Kind 1 as well. Stable uses it: it is when the feed was first added to the database.
```
cka-y: [question] are these comments refactoring really necessary?

functions-python/tasks_executor/src/tasks/seal_of_reliability/backfill/seal_backfill.py
```diff
@@ -0,0 +1,785 @@
...
+criterion of the run. So re-running an interrupted run finishes it, and re-running a finished
+one writes nothing.
+
+On each marched day, a criterion is evaluated only where `seal_criterion_snapshot` holds no
+verdict of its own for it. A row already holding PASS, FAIL or NOT_APPLICABLE supplies the
+day's `observed_status` and the evaluator is not called; a missing row, or one holding
+UNKNOWN or NEVER_EVALUATED, is no verdict, so the evaluator is called. Grace, probation and
+the streak are worked out from whichever status was used, day by day, as always.
+
+Two limits, argued in misc/AI/seal_backfill_algorithm_1763.md: Official and Stable have no
```
davidgamez: This comment is about a file that doesn't exist the repo

functions-python/tasks_executor/src/tasks/seal_of_reliability/backfill/seal_backfill.py
```diff
@@ -0,0 +1,785 @@
...
+    _validate_requested_feed_ids,
+    _write_snapshot_rows,
+    is_partial_run,
+)
+from tasks.seal_of_reliability.state_machine import SealCriterionState, transition
+
+logger = logging.getLogger(__name__)
+
+# Days rather than months, so the arithmetic needs no calendar library: 365 is #1763's
+# "12 months".
```
davidgamez: This comment is unnecessary

functions-python/tasks_executor/src/tasks/seal_of_reliability/backfill/seal_backfill_orchestrator.py
```diff
@@ -0,0 +1,193 @@
...
+has a reconstruction written over it.
+
+Payload (all optional)::
+
+    {
+        "dry_run": bool,                  # default True
+        "batch_size": int,                # default 100
+        "start_date": str | None,         # ISO date, default end_date - days_back
+        "end_date": str | None,           # ISO date, default yesterday UTC
+        "days_back": int,                 # default 365
```
davidgamez, line 34: [suggestion]: I would reduce the default number of days back just to avoid the full year run on unintended executions.

functions-python/tasks_executor/src/tasks/seal_of_reliability/backfill/seal_backfill_orchestrator.py
```diff
@@ -0,0 +1,193 @@
...
+        "criteria": [str] | None,         # default None (every implemented criterion)
+        "limit": int | None,              # cap total feeds considered, default None
+        "stable_feed_ids": [str] | None,  # restrict eligibility to these ids, default None
+        "only_missing": bool,             # default True
+        "snapshot_mode": str,             # final | all | none, default all
+        "resume_from_snapshot": bool,     # default False
+        "deadline_seconds": int,          # default 7200 (2h)
```
davidgamez, line 43: The maximum timeout for an HTTP-triggered Cloud Run function is 3600; I don't think 7200 will ever be reached unless I'm interpreting this parameter wrong.

functions-python/tasks_executor/src/tasks/seal_of_reliability/backfill/seal_backfill.py
jcpitre: Removed

functions-python/tasks_executor/src/tasks/seal_of_reliability/backfill/seal_backfill.py
jcpitre: Removed

functions-python/tasks_executor/src/tasks/seal_of_reliability/backfill/seal_backfill_orchestrator.py
jcpitre: The timeout is handled by the monitor. The monitor is given the starting time of the whole process by the orchestrator. If after 2 hours passed on the clock it did not collect all the workers output, it will fail.

functions-python/tasks_executor/src/tasks/seal_of_reliability/backfill/seal_backfill_orchestrator.py
jcpitre: What number do you suggest?

functions-python/tasks_executor/src/tasks/seal_of_reliability/backfill/seal_backfill_orchestrator.py
jcpitre: Added a comment.

davidgamez: Got it, Thanks!

davidgamez (approved):

PR #1819, opened 2026-08-27
feat: available + compliant criteria evaluation and seal status logic update

### LINKED ISSUE #1784: Seal of reliability: Implement other criteria evaluation

cka-y:
**Summary:**

Completes the Seal of Reliability criteria set for #1784 by adding the last two evaluators, **Available** and **Compliant**, and makes the feed-level outcome say *why* a feed does not hold the seal. Builds on #1817 (Official / Stable / Fresh) and follows the algorithm in #1761. Fresh / continuous coverage remains out of scope and is tracked by #1782.

Three changes:

1. **Available** (`available.py`) — the latest `gtfs_feed_availability_check` in a rolling 24h window has `success = TRUE`. A window with no check at all is `unknown`, not a failure: it means we did not look, not that the feed was down. 14-day grace period, serves probation.

2. **Compliant** (`compliant.py`) — `total_error = 0` on the **latest validation report of the feed's latest dataset**. Scoped to the latest dataset rather than to the feed: a verdict on a superseded dataset does not describe what we are serving today. A latest dataset with no report yet (unvalidated, or validation lagging publication) is `unknown`, which freezes the criterion at its last confirmed verdict instead of failing it. 30-day grace period, serves probation.

3. **Feed-level `seal_status`** — `has_seal` is a boolean and cannot distinguish *judged and did not qualify* from *cannot be judged yet*. `roll_up_seal_status` in `shared/common/seal_criteria.py` now derives a four-value status: `granted`, `not_granted`, `unknown` (at least one criterion has never produced a verdict), `never_evaluated` (none ever has). It is **derived, not stored** — a pure function of `confirmed_status` plus probation on rows both the job and the API already hold — so there is **no schema change** in this PR. Exposed on `GET /v1/gtfs_feeds/{id}/reliability` as a new optional field.

Every criterion evaluation also now emits a log line (feed, criterion, observed, confirmed, previous confirmed, phase, reason), plus one per feed for the roll-up. The run report gains `seal_status` per feed and a `seal_status_counts` summary.

**Expected behavior:**

- A feed meeting every criterion is `granted` and holds the seal, as before.
- A feed whose criteria were all judged but does not qualify is `not_granted`, `has_seal: false`.
- A feed with any criterion that has never produced a verdict is `unknown`, `has_seal: false` — it is *not* denied the seal, it just cannot be decided yet. On a fresh environment most feeds sit here until availability checks and validation reports exist.
- A transient upstream outage does **not** move the seal: a criterion that already has a stored verdict keeps it through an `unknown` run, so one bad night for the availability or validation job cannot blank the seal across the catalogue.
- `GET /v1/gtfs_feeds/{id}/reliability` returns the new `seal_status` alongside the existing `has_seal` and the six criteria.


api/.openapi-generator/FILES
```diff
@@ -52,25 +52,3 @@ src/feeds_gen/models/search_feeds200_response.py
 src/feeds_gen/models/source_info.py
 src/feeds_gen/models/validation_report.py
 src/feeds_gen/security_api.py
-src/user_service/impl/__init__.py
```
davidgamez, line 59: Please run the `./scripts/api-operations-update-schema.sh` script to add back these files.

api/src/shared/common/seal_criteria.py
```diff
@@ -111,10 +125,49 @@ class CriterionPhase(str, Enum):
 )
 
 
+def roll_up_seal_status(
+    criteria: Iterable[Tuple[CriterionStatus, bool]],
+) -> SealStatus:
+    """The feed-level seal outcome from its criteria.
+    * every criterion NEVER_EVALUATED -> seal NEVER_EVALUATED.
```
cka-y: My logic could be wrong but i thought if every criterion is never evaluated then we can conclude that the overall seal is never evaluated, but if some have been evaluated (and pass), then the overall seal is unknown because some has never been evaluated. But i agree it's a bit confusing and both scenario could result in NEVER_EVALUATED for simplicity.

api/src/shared/common/seal_criteria.py
```diff
@@ -111,10 +125,49 @@ class CriterionPhase(str, Enum):
...
+    in_scope = [
+        (status, on_probation) for status, on_probation in criteria if status is not CriterionStatus.NOT_APPLICABLE
+    ]
+    if not in_scope:
+        # Every criterion is NOT_APPLICABLE, so there is nothing left to judge the feed by.
+        return SealStatus.NEVER_EVALUATED
+
+    unjudged = sum(1 for status, _ in in_scope if status is CriterionStatus.NEVER_EVALUATED)
+    if unjudged == len(in_scope):
+        # Every criterion that isn't NOT_APPLICABLE is NEVER_EVALUATED
```
jcpitre:
```suggestion
        # All criteria are either NOT_APPLICABLE or NEVER_EVALUATED: nothing to judge the feed by.
```

api/src/shared/common/seal_criteria.py
```diff
@@ -111,10 +125,49 @@ class CriterionPhase(str, Enum):
...
+        return SealStatus.NEVER_EVALUATED
+
+    # One criterion is enough to deny the seal, so this is decidable even with the rest unjudged.
+    # A pass on probation denies it too: probation withholds the criterion whatever its status.
+    denied = any(
+        status is CriterionStatus.FAIL or (status is CriterionStatus.PASS and on_probation)
```
jcpitre: I would replace the variable `status` by `confirmed_status` to help understanding.

api/.openapi-generator/FILES
jcpitre, line 59: Is this supposed to be removed?

functions-python/tasks_executor/src/tasks/seal_of_reliability/evaluators/compliant.py
```diff
@@ -0,0 +1,57 @@
...
+"""Compliant criterion: the latest dataset validates with no errors."""
+
+from typing import Tuple
+
+from shared.common.seal_criteria import CriterionStatus, SealCriterionName
+from tasks.seal_of_reliability.context import FeedSealContext
+from tasks.seal_of_reliability.evaluators.base import CriterionEvaluator
+
+
+class CompliantEvaluator(CriterionEvaluator):
```
davidgamez, line 25: Does this evaluator need to be time-aware? How will the backfill (not part of this issue) work if we only evaluate the latest dataset report?

cka-y: it is time aware. it takes the latest dataset as of the value of `now`. Shoudl i rename the variables for better clarity?

davidgamez (approved):

PR #1816, opened 2026-08-26
feat: api continuous coverage endpoints

cka-y:
**Summary:**

Closes MobilityData/product-tasks#215

Adds `GET /v1/gtfs_feeds/{id}/continuous_coverage`, which returns a GTFS feed's dataset history (newest to oldest, paginated) for the "continuous coverage" criterion of the Seal of Reliability.

**Expected behavior:**

- `GET /v1/gtfs_feeds/{id}/continuous_coverage` returns 200 with paginated `items[]` plus the `latest_*` summary fields for the feed's current dataset.
- `downloaded_after`/`downloaded_before` that are malformed or out of order return 400.
- An unknown feed id returns 404.
- A feed with no datasets still returns a valid response: `latest_files` lists all three tracked files as absent, and every other `latest_*` field is null.

**Testing tips:**

- `./scripts/api-tests.sh --test_file api/tests/unittest/models/test_continuous_coverage.py`
- `./scripts/api-tests.sh --test_file api/tests/unittest/models/test_gtfs_feed_continuous_coverage_impl.py`
- `./scripts/api-tests.sh --test_file api/tests/unittest/test_feeds.py`
- Locally: `scripts/api-start.sh`, then hit `GET localhost:8080/v1/gtfs_feeds/<id>/continuous_coverage` for a feed with several datasets; vary `limit`/`offset`/`downloaded_after`/`downloaded_before` and confirm `latest_*` stays constant across pages while `items[]` changes.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Run the unit tests with `./scripts/api-tests.sh` to make sure you didn't break anything
- [x] Add or update any needed documentation to the repo
- [x] Format the title like "feat: [new feature short description]". Title must follow the Conventional Commit Specification(https://www.conventionalcommits.org/en/v1.0.0/).
- [x] Linked all relevant issues
- [ ] Include screenshot(s) showing how this pull request works and fixes the issue(s)

PR #1824, opened 2026-09-02
feat: fresh continuous coverage criterion impl

### LINKED ISSUE #1782: Seal of Reliability: Implement the Fresh-continuous-coverage criterion

cka-y:
**Summary:**
Closes #1782

Implements the sixth and last Seal of Reliability criterion, `fresh_continuous`: successive datasets cover service with no uncovered day between them. With this the evaluator registry covers `seal_criterion_name` exactly, so a partial run is now only ever one an operator asked for.

- **`evaluators/fresh_continuous.py`** (new) — the criterion, as three ordered steps (chart below).
- **`context.py`** — `ClosestDataset` becomes `DatasetCoverage`, now carrying both coverage windows plus `has_calendar_data`, and the context gains `previous_dataset`. `_load_closest_datasets` becomes `_load_recent_datasets`: one `ROW_NUMBER()` query per batch returning each feed's two most recent datasets, still bounded by the run's `now` so replays and backfills stay honest.
- **`shared/common/continuous_coverage.py`** — added `FEED_INFO_FILE` and `CALENDAR_FILES`; `COVERAGE_FILES` is now composed from them, so the tuple the continuous-coverage endpoint (#1816) serves is byte-for-byte unchanged. `MAX_COVERAGE_WINDOW`, `overlap_and_gap` and `within_max_coverage_window` are reused as-is, so the criterion and the read endpoint cannot disagree about what "successive datasets overlap" means.

No schema change: `seal_criterion_name` already declared `fresh_continuous`, and its grace period (none) and probation period (180 days) were already in the policy maps.

**Expected behavior:**

Three steps, in order. The first two are settled on the closest dataset alone, which is why a feed can fail on a single dataset.

```mermaid
flowchart TD
    A["closest dataset<br/>as of now"] -->|none| U1(["UNKNOWN<br/>no dataset"])
    A --> B{"window?<br/>feed_info range,<br/>else service window"}

    B -->|"none, has calendar file"| U2(["UNKNOWN<br/>not derived yet"])
    B -->|"none, no calendar file"| F1(["FAIL<br/>nothing published"])
    B -->|"longer than 730 days"| F2(["FAIL<br/>placeholder calendar"])
    B -->|ok| C{"previous<br/>dataset?"}

    C -->|none| P1(["PASS<br/>no boundary yet"])
    C --> D{"both datasets declare<br/>a feed_info range?"}

    D -->|"yes, no gap"| P2(["PASS"])
    D -->|"yes, gap"| E{"validated service<br/>windows"}
    D -->|no| E

    E -->|no gap| P3(["PASS<br/>calendar excuses it"])
    E -->|gap| F3(["FAIL"])
    E -->|"missing, no calendar file"| F4(["FAIL<br/>producer's omission"])
    E -->|"missing, has calendar file"| U3(["UNKNOWN<br/>ours to derive"])
```

Two things in there are decisions rather than mechanics:

**The declared range outranks the calendar, and a gap has to show in both to fail.** `feed_info.txt` is the service period the producer intends, so where both datasets declare one it decides; the validated service window is what excuses a declared gap, and is the whole answer when they don't both declare one.

**A missing window is the producer's answer, not a missing one** (per the Slack thread of 2026-08-27, since neither #1782 nor the "how it's calculated" page covers it). If a dataset shipped no `calendar.txt`/`calendar_dates.txt` at all, that is a FAIL — asking two consecutive datasets to carry calendar data is not a high bar. It stays UNKNOWN only when the file *is* there and we have not derived a window from it, which is our lag rather than theirs. That is the `has_calendar_data` flag, and it is why the context loads file presence alongside the windows rather than inferring one from the other.

Also worth flagging for review, since neither is in the issue:

- **Seasonal feeds get no exemption**, unlike `fresh_coverage`. A between-season gap is a real gap; the way to show it is intentional is to keep one of the two windows continuous across it.
- **The 730-day threshold applies to the closest dataset only**, not to the previous one, and it is not re-applied inside the calendar branch.

### DIFF
--- a/functions-python/tasks_executor/src/tasks/seal_of_reliability/evaluators/fresh_continuous.py
+++ b/functions-python/tasks_executor/src/tasks/seal_of_reliability/evaluators/fresh_continuous.py
@@ -1,0 +1,156 @@
+#
+#   MobilityData 2026
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+"""Fresh (continuous coverage) criterion: successive datasets cover service with no gap."""
+
+from datetime import date
+from typing import Optional, Sequence, Tuple
+
+from shared.common.continuous_coverage import (
+    CALENDAR_FILES,
+    FEED_INFO_FILE,
+    MAX_COVERAGE_WINDOW,
+    as_date,
+    overlap_and_gap,
+    window_days,
+    within_max_coverage_window,
+)
+from shared.common.seal_criteria import CriterionStatus, SealCriterionName
+from tasks.seal_of_reliability.context import DatasetCoverage, FeedSealContext
+from tasks.seal_of_reliability.evaluators.base import CriterionEvaluator
+
+Window = Tuple[date, date]
+
+CALENDAR_FILE_LIST = " or ".join(CALENDAR_FILES)
+
+
+def _window(start, end) -> Optional[Window]:
+    """Two plain dates, or None when the window is incomplete or inverted."""
+    start, end = as_date(start), as_date(end)
+    return (start, end) if window_days(start, end) is not None else None
+
+
+def _declared_window(dataset: DatasetCoverage) -> Optional[Window]:
+    return _window(dataset.feed_info_start, dataset.feed_info_end)
+
+
+def _service_window(dataset: DatasetCoverage) -> Optional[Window]:
+    return _window(dataset.service_date_range_start, dataset.service_date_range_end)
+
+
+def _span(window: Window) -> str:
+    return f"{window[0].isoformat()} to {window[1].isoformat()}"
+
+
+def _calendar_verdict(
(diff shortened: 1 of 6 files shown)


functions-python/tasks_executor/src/tasks/seal_of_reliability/evaluators/fresh_continuous.py
```diff
@@ -0,0 +1,162 @@
...
+            )
+        if within_max_coverage_window(*window) is False:
+            return (
+                CriterionStatus.FAIL,
+                f"dataset {newer.dataset_id} covers {_span(window)}, longer than the "
+                f"{MAX_COVERAGE_WINDOW.days}-day maximum coverage window",
+            )
+        # 2. nothing published before it, so no boundary to judge
+        older = ctx.previous_dataset
+        if older is None:
```
davidgamez, line 140: 👍

davidgamez (approved):
LGTM! 🥇

tzujenchanmbd:
Expected behavior flowchart 🦑 lgtm

PR #1829, opened 2026-09-09
feat: added latest failure and reformated latest state

cka-y:
**Summary:**
Closes https://github.com/MobilityData/product-tasks/issues/219

`GET /v1/gtfs_feeds/{id}/continuous_coverage` reported the current state as nine flattened `latest_*` fields. Those are replaced by two objects, `latest_state` and `latest_failure`, both `$ref`-ing the existing `GtfsFeedContinuousCoverage` schema so each has the structure of an `items[]` entry.

- `latest_state` — the feed's latest dataset measured against the one before it.
- `latest_failure` — the same measurement at the `fresh_continuous` criterion's `last_observed_failure_at`, so the endpoint cannot disagree with the seal's verdict. `null` when the criterion has never been observed failing.

Each state names its own dataset plus its `previous_dataset_id`: at most four datasets across both, shared when the current state is itself the failure. `items`, `total`, `offset`, `limit` and the query params are unchanged.

**Breaking change** — removed: `latest_files`, `latest_coverage_window`, `latest_coverage_window_source`, `latest_within_max_coverage_window`, `latest_service_window`, `latest_feed_info_window`, `latest_feed_info_matches`, `latest_overlap_days`, `latest_gap_days`. Same data, now under `latest_state`.

Stubs regenerated with `scripts/api-gen.sh`, `scripts/api-operations-update-schema.sh` and `scripts/api-operations-gen.sh`.

**Expected behavior:**

Currently passing feed: `latest_state` is the latest dataset, `latest_failure` is `null` or an older boundary. Currently failing feed: both describe the same dataset, with `latest_failure.is_latest` true. Feed with no datasets: both `null`.

### DIFF
--- a/api/src/feeds/impl/feeds_api_impl.py
+++ b/api/src/feeds/impl/feeds_api_impl.py
@@ -21,11 +21,9 @@ from feeds_gen.models.gtfs_dataset import GtfsDataset
 from feeds_gen.models.gtfs_feed import GtfsFeed
 from feeds_gen.models.gtfs_feed_availability_response import GtfsFeedAvailabilityResponse
 from feeds_gen.models.gtfs_feed_continuous_coverage import GtfsFeedContinuousCoverage
-from feeds_gen.models.gtfs_feed_continuous_coverage_file import GtfsFeedContinuousCoverageFile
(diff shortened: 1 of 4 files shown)


davidgamez (approved):
LGTM!

PR #1830, opened 2026-09-09
feat: supporting is_seasonal manual set to TDG, JBDA, ODPT

### LINKED ISSUE #1798: is_seasonal: adding to TDG, JBDA, ODPT
We want to be able to use what's in #1774 for feeds that are automatically imported into the Mobility Database, including from TDG, ODPT  and JBDA. 

Since there are seasonal feeds from these sources, we want to be able to set is_seasonal=true for these feeds so the 7 day rolling window is omitted from the Seal calculation.

davidgamez:
**Summary:**

Makes `seasonal` an operator-owned, tri-state field so a manual setting survives every automated import and update.

- **Operations API:** replaced `operational_status_action` (enum including `no_change`) with `operational_status` (`wip|published|unpublished`) on the GTFS and GTFS-RT update requests. `operational_status` and `seasonal` are now tri-state: omit to preserve the stored value, send a value to set it. A feed GET response can be sent straight back as an update with no change detected.
- **Operations impl:** `operational_status` is now exposed via `from_orm`, so it is visible to change detection. `_PRESERVE_WHEN_OMITTED_FIELDS` backfills omitted tri-state fields from current values to avoid phantom diffs, and the diff becomes the single write gate - the `operational_status_action` special case in `_populate_feed_values` and `_update_feed` is gone. `to_orm` only writes `seasonal` / `operational_status` when explicitly provided.
- **Importers (TDG, ODPT, JBDA):** added `OPERATOR_OWNED_FEED_COLUMNS = ("seasonal",)`. No upstream source carries a seasonality signal, so importers never assign it. TDG additionally carries `seasonal` across the delete/recreate path when a feed's `data_type` flips.
- **Seal of Reliability docs:** `CriterionContribution.status` is redocumented as the debounced verdict; the at-risk state is now `pass` with `in_grace_period` true rather than `fail`.

Note: `operational_status_action` → `operational_status` is a breaking spec change.

**Expected behavior:**

- Update request omitting `seasonal` or `operational_status` → stored values unchanged, 204 (no needless materialized view refresh or web revalidation).
- Update request sending either explicitly → value written, 200.
- Re-running any importer over a feed manually marked seasonal → flag preserved, including a TDG `data_type` change.

**Testing tips:**

```bash
scripts/api-tests.sh --folder functions-python/operations_api

### CODE REVIEW

**Comment by jcpitre** [comment]
Out of scope for this PR, but something keeps adding blank lines to OperationsAPI.yaml:

https://github.com/MobilityData/mobility-feed-api/blob/f4a31ad74e282b20f3dabd34882d6b225ffdbd5f/docs/OperationsAPI.yaml#L2914-L2921
There's other locations in the file where blank lines are added.

**Comment by davidgamez** [comment]
> Out of scope for this PR, but something keeps adding blank lines to OperationsAPI.yaml:
> 
> https://github.com/MobilityData/mobility-feed-api/blob/f4a31ad74e282b20f3dabd34882d6b225ffdbd5f/docs/OperationsAPI.yaml#L2914-L2921
> 
> 
> There's other locations in the file where blank lines are added.

Fixed

File: docs/OperationsAPI.yaml [hunk 1]
```diff
@@ -2927,28 +2794,23 @@ components:
           items:
             type: string
             example: "mdb-20"
-        operational_status_action:
```
jcpitre, line 2930: I think some tests are still using operational_status_action.
See https://github.com/MobilityData/mobility-feed-api/blob/09c76a27405a2d8ed952e24f7f71d67c3036a11d/scripts/api-operations-surface.sh#L222

davidgamez: That test is expecting a 400 error code

jcpitre (approved):

docs/OperationsAPI.yaml
jcpitre: But won't it accept 200 also?

PR #1839, opened 2026-09-14
feat: validation reports endpoint

cka-y:
**Summary:**
Closes https://github.com/MobilityData/product-tasks/issues/211
Adds `GET /v1/gtfs_feeds/{id}/validation_reports` - the feed's validation history, one entry per
dataset, carrying the counts of that dataset's most recent report and the notice codes behind them.

The `notice` table (notice code, severity, count per report) was already populated but was not
reachable through the API. Until now a client could see *that* a dataset had 10 errors but not
*which* errors, and there was no way to see validation over time without walking datasets one by
one. This gives the UI what it needs to display a feed's error state.

Deliberately not a Seal of Reliability endpoint: it reports what the validator found, with no
criterion, verdict or seal vocabulary in the response. The `compliant` criterion happens to read
the same reports, but nothing here is derived from it.

- One entry per dataset, the most recent report by `validated_at` (ties broken by report id).
- `latest` is the entry for the feed's current dataset, unaffected by paging and filters, so the
  current error state comes back in the same call as the history.
- Filters: `validated_after`, `validated_before`, `min_errors`, `min_warnings`, and `severity`
  (repeatable). `severity` narrows the `notices` array only - the counts still report everything
  found, so filtering to ERROR cannot make a feed look cleaner than it is.
- Notices are ordered errors first, then by count, so the head of the list is what breaks the feed.

Additive only - no existing endpoint or response changes.

**Expected behavior:**

`GET /v1/gtfs_feeds/mdb-1/validation_reports?limit=1`

```json
{
  "feed_id": "mdb-1",
  "latest": { "dataset_id": "dataset-2", "is_latest": true, "...": "same shape as an item" },
  "total": 2,
  "offset": 0,
  "limit": 1,
  "items": [
    {
      "dataset_id": "dataset-2",
      "is_latest": true,
      "validated_at": "2023-02-01T10:10:10Z",
      "validator_version": "2.0.0",
      "total_error": 10,
      "total_warning": 20,
      "total_info": 30,
      "unique_error_count": 1,
      "unique_warning_count": 2,
      "unique_info_count": 3,
      "url_json": "https://.../report.json",
      "url_html": "https://.../report.html",
      "notices": [
        { "code": "foreign_key_violation", "severity": "ERROR", "total": 10 },
        { "code": "inconsistent_agency_timezone", "severity": "WARNING", "total": 1 }
      ]
    }
  ]
}
```

Three queries per response regardless of page size: the ranked report per dataset, one batched
notice fetch for the whole page, and one for `latest`.

**Testing tips:**

`GET /v1/gtfs_feeds/{id}/validation_reports` on a feed with several datasets, and check:

- One entry per dataset, not one per validation report - a dataset with several reports (different
  validator versions) appears once, with its most recent.
- `items` are ordered by `validated_at`, newest first, and exactly one entry has `is_latest: true`.
- `latest` matches the entry flagged `is_latest`, and stays populated when a filter empties
  `items` (e.g. `min_errors=1000000`).
- `notices` carries the codes behind the counts, errors before warnings before info.
- `severity=ERROR` returns only ERROR notices while `total_warning` on the same entry is unchanged.
- `min_errors=1` lists only datasets that failed validation; `min_errors=0` changes nothing.
- `validated_after` / `validated_before` narrow the history; an inverted range returns 422, as does
  a malformed date.
- An unknown feed id returns 404.

Note: the endpoint documents **422** for invalid parameters, which is what the API actually returns.
`/availability` and `/continuous_coverage` document 400 for the same case while also returning 422 -
a pre-existing inconsistency left alone here rather than widened.


davidgamez (approved):
LGTM!

PR #1841, opened 2026-09-14
fix: probation response on criterion failure (search endpoint)

cka-y:
**Summary:**

Follow-up to #1840, which fixed the same rule everywhere except search.

`/v1/search` does not read the seal tables — it reads the precomputed `seal_latest_probation_start` column from the `feedsearch` materialized view, so it did not pick up the fix in #1840 and kept reporting a criterion as serving probation when its most recent check observed a failure. That left the two paths disagreeing about the same feed: `on_probation: true` from search, `false` from the feed-detail and reliability-report endpoints.

One line in the view's probation aggregate closes the gap:

```sql
MAX(SealCriterion.probation_start) FILTER (
    WHERE SealCriterion.criterion NOT IN ('official', 'stable')
      AND SealCriterion.observed_status <> 'fail'
) AS seal_latest_probation_start
```

This now mirrors `is_serving_probation()` in `api/src/shared/common/seal_criteria.py`, which the other endpoints go through, and the comment above the join says so, so the two roll-ups are findable from each other.

No new changeset: the enclosing changeset in `liquibase/materialized_views/materialized_views.xml` is `runOnChange="true"`, so editing the SQL is enough for liquibase to drop and recreate `feedsearch` on the next migration. No schema change and no change to the view's column set, so no model regeneration either.

**Expected behavior:**

For a feed with a non-exempt criterion (`available`, `compliant`, `fresh_coverage`, `fresh_continuous`) carrying a `probation_start` and `observed_status = fail`, `reliability_seal` in a search result changes:

| Field | Before | After |
| --- | --- | --- |
| `on_probation` | `true` | `false` |
| `probation_ends_at` | a date | `null` |
| `has_seal` | unchanged | unchanged |

Since the roll-up is a `MAX`, a feed with several probationary criteria now reports the latest probation start *among those not currently failing* — which can be an older date than before, and is the intended reading: the failing criterion is failing, not recovering.

Everything else is untouched. A criterion passing while on probation still rolls up with its countdown, and `official` / `stable` stay exempt.

After this merges, search and the feed-detail / reliability-report endpoints agree on probation for every feed, and the "known divergence" noted in #1840 is resolved.

**Testing tips:**

```bash
./scripts/api-tests.sh --folder api
```

Two tests were added to `api/tests/integration/test_search_api.py`, which exercise the real view through the search endpoint (`_feed_with_seal` grew an `extra_criteria` argument to seed `seal_criterion` rows):

- `test_search_reports_probation_from_the_view` — a criterion passing on probation still reports `on_probation: true` with a countdown.
- `test_search_ignores_probation_for_a_criterion_observed_failing` — the case this PR fixes.

The second test fails against the pre-fix view and passes against the new one, so it is a real regression guard rather than a restatement of the code.

Note for local testing: search reads a materialized view, so the view has to be rebuilt before the tests see the change. A full `liquibase update` against the local test DB may fail for unrelated reasons; applying just this file does the same thing the changeset does:

```bash
docker cp liquibase/materialized_views/feed_search.sql database_test:/tmp/feed_search.sql
```

then `DROP MATERIALIZED VIEW IF EXISTS feedsearch;` followed by `\i /tmp/feed_search.sql` in `psql` against `MobilityDatabaseTest`.

On deploy, the migration recreates the view, but the data behind it is only as fresh as the next `REFRESH` — worth a look at the `refresh_feedsearch_view` cadence if the corrected values are wanted immediately.


liquibase/materialized_views/feed_search.sql
```diff
@@ -127,6 +130,7 @@ LEFT JOIN (
         MAX(SealCriterion.evaluated_at) AS seal_evaluated_at,
         MAX(SealCriterion.probation_start) FILTER (
             WHERE SealCriterion.criterion NOT IN ('official', 'stable')
+              AND SealCriterion.observed_status <> 'fail'
```
davidgamez, line 133: 👍

davidgamez (approved):

